PR #5, opened 2018-07-16
Use caf::variant, remove broker::detail::variant

Neverlord:
We already have CAF as primary dependency, so it makes little sense to ship yet another custom variant implementation. Aside from reducing code duplication, CAF's variant is also well-tested by now and fully C++17 compatible.

@jsiwek I'm not sure if the GitHub version is synchronized back upstream. Feel free to redirect me someplace else if GitHub isn't the right place for filing patches to Broker. 🙂


mavam:
Always good to see commits that throw out a bunch of code.

This PR depends on a not-yet-existant commit in Bro that performs the corresponding refactoring, such as replacing `broker::get` with `caf::get`.

jsiwek:
Thanks, I've conservatively added back in some compat/wrapper functions -- we've already tagged a Broker release a while ago (maybe jumped the gun a bit) and know people are using it, so nice to not break existing code unless necessary.

(updating the Bro side of things wasn't needed with the wrapper functions, but I went ahead anyway)

Neverlord:
> I've conservatively added back in some compat/wrapper functions

Yeah, that makes sense. I didn't know that users already rely on the API, thanks for completing the task!

PR #7, opened 2018-08-20
Allow users to spawn the core actor without an endpoint

Neverlord:
Refactor the clock component to allow users to pass around a clock pointer instead of a full endpoint and move the core actor to the public namespace.


jsiwek:
Merged, but I ported back some of the original `clock::advance_time()` as it's performance-sensitive and the changes did regress in that aspect (test case measured ~50% worse w/ culprit being the unnecessary `std::unordered_set` ctor along some paths that can rather early-out).

PR #10, opened 2018-09-05
Make blocking flags available to Python wrapper

mavam:
The C++ API offers two flags that weren't available in the Python bindings. This PR adds them.


jsiwek:
Thanks.

PR #14, opened 2018-09-12
Remove unused warnings and variables

mavam:
This PR removes warnings about unused warnings: by (1) silencing unused arguments in non-debug builds, and (2) actually removing an unused variable.


jsiwek:
Thanks.

PR #45, opened 2019-06-07
Add doc string for enum_value for zeek/broker#44

### LINKED ISSUE #44: Missing documentation of enum_value
I noticed in the documentation that the [types section](https://docs.zeek.org/projects/broker/en/stable/data.html#types) of the broker lacks a section about the `enum_value` type.

The header `broker/data.hh` clearly supports `enum_value` as a `data_variant` so it should be included in the docs.

I'd update the file `doc/data.rst` with a sentence or two that describes the type like so:

> An `enum_value` wraps `enum` types defined by zeek and is a type alias for `std::string`.

However, I am not sure that this accurate, given that I just learned about its existence.

NSkelsey:
This closes #44

### DIFF
--- a/doc/data.rst
+++ b/doc/data.rst
@@ -95,6 +95,10 @@ Broker directly supports |std_string|_ as one possible type of ``data``.
 .. |std_string| replace:: ``std::string``
 .. _std_string: http://en.cppreference.com/w/cpp/string/basic_string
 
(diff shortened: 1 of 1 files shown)


jsiwek:
Thanks, I also extended it a bit with the description from `broker/enum_value.hh`

PR #50, opened 2019-08-15
Revert 90a7ab3 and adjust base_fixture to CAF 0.17

Neverlord:
This PR reverts the changes from 90a7ab3 to `integration.cc` and `master.cc`:
- the `integration` unit test fails because of a bug in CAF 0.17: https://github.com/actor-framework/actor-framework/issues/880
- the `master` test runs fine when setting the parameter `middleman.workers` to 0 to restore the behavior on CAF < 0.17

@jsiwek how is your policy regarding `CHANGES` and `VERSION` here? Should I revert the changes from 90a7ab3 in these files as well or add new entries?

The `integration` test is broken on 0.17, but I think we should not deploy a workaround for the broken release and instead simply release CAF 0.17.1 soon and skip support for 0.17 with Broker.


jsiwek:
Thanks for looking into these.

> * the `master` test runs fine when setting the parameter `middleman.workers` to 0 to restore the behavior on CAF < 0.17

Is there anything more general you can think we should be doing to make the Broker tests less sensitive to changes in CAF ?

> @jsiwek how is your policy regarding `CHANGES` and `VERSION` here? Should I revert the changes from [90a7ab3](https://github.com/zeek/broker/commit/90a7ab3ebcc041ff89b378ffdb42f69117707d86) in these files as well or add new entries?

Don't revert those files -- `CHANGES` will just be an accurate history of what happened, and then `VERSION` is something that gets updated by the person performing the merge into master (we have a script to auto-update it).

> The `integration` test is broken on 0.17, but I think we should not deploy a workaround for the broken release and instead simply release CAF 0.17.1 soon and skip support for 0.17 with Broker.

Yeah, once a CAF 0.17.1 is out, I don't expect the `integration` test to have any workarounds just to support 0.17.0, too.

To clarify: the bug only effects deterministic setups like our test suite and not realtime use-cases ?  If so, then it's not too high priority (in terms of Broker), so fine with leaving the test case with a workaround for however long you need to prepare CAF 0.17.1.

Neverlord:
> Is there anything more general you can think we should be doing to make the Broker tests less sensitive to changes in CAF ?

The issue is that Broker hand-rolls functionality that's already in CAF. Here, the `base_fixture` in Broker is essentially what CAF provides as `test_coordinator_fixture`. The CAF fixture added that option upstream, so we wouldn't have noticed a change in Broker if we had used the CAF fixture as base.

> To clarify: the bug only effects deterministic setups like our test suite and not realtime use-cases ?

Yes, this bug can only happen if you advance everything manually in a test setup.

> fine with leaving the test case with a workaround for however long you need to prepare CAF 0.17.1.

👍

There's nothing holding back an 0.17.1 release. I'd only like to finish working on the recording/generating feature in `topic/neverlord/recording` to see if anything else comes up.

Neverlord:
@jsiwek `0.17.1` just released. This PR should build on the new tag.

jsiwek:
When updating to CAF 0.17.1 and merging this, I see test failures:

```
The following tests FAILED:
	  2 - core (Child aborted)
	  4 - status_subscriber (Child aborted)
	  6 - master (Child aborted)
	  7 - publisher (Child aborted)
	 11 - subscriber (Child aborted)
```

All fail for same reason shown here:

```
ctest -V -R core

...

2: terminate called after throwing an instance of 'std::runtime_error'
2:   what():  Cannot start OpenSSL module without default backend.
1/1 Test #2: core .............................***Exception: Child aborted  0.11 sec
```

They pass w/ just the original commit in this PR, 8cbd7189490815e7bba693447625ca4f5942075f, but break with the changes in 7b31368587223740c66188132b1dae0dd85c364a.

Neverlord:
Should be fixed now.

jsiwek:
Thanks, everything is merged, but I also noticed that we're effected by this memory leak fix that's an issue in 0.17.1:

https://github.com/actor-framework/actor-framework/commit/f3aa0dfd63a68534f255bcb042641b7888f4e769

In the interest of just getting this PR moved along, I just cherry-picked that fix into the 0.17.1 CAF fork we embed in Zeek/Broker by default, but probably nice to get a 0.17.2 patch release which includes it sometime soon-ish.

PR #56, opened 2019-09-20
Add new cluster benchmarking tool

Neverlord:
The new tool comes with a `README.md` as well as a zipped example setup to get people started.


jsiwek:
Thanks, looks neat.  Some questions about the cluster config file I didn't understand:

* "A node cannot send and receive at the same time"

In a real cluster, seems like that's actually a common case to expect, so why is not allowed for the benchmark ?

And even though it's enforced that a node in the config file doesn't use both an input and an output option, it seems like a node may be implicitly violating that rule anyway by setting both `generator-file` and `topics` ?  I.e. it's producing output, but also subscribing to some topics and thus potentially receiving messages from some other node in the network even though it's not also using `num-inputs`.

* "At least one node must receive data."

Seems that's enforced by looking for at least one node with `num-inputs` ?

But wonder that a common usage would be "just run the benchmark on the full dataset".  In that case, seems a bit awkward to have to explicitly set the expected receive number instead of an option that tells the node "expect to receive unknown amount of data, but upstream publishers will notify you when they're done sending".  That way, you still have a termination condition for the node.  What do you think?

* How's data recorded by a real Zeek cluster supposed to be used to set up a benchmark?

This question maybe just relates to the first two bullet points causing friction with the way I first expected I might benchmark using data produced by a real Zeek cluster.  E.g. say I have data recorded from a 4 node cluster (a "logger", "worker", "manager", and "proxy"), my first thought was that I should be able to set up a single config file for the benchmark to effectively simulate that cluster environment, replay the data through it, and see how long it takes.  Or is there a different approach you're thinking of ?

Because of the input/output restrictions it kind of seems like I'd actually need 4 or more different config files to test all the recorded data.  E.g. for each 4 nodes, one node uses its associated `generator-file` and all others are set to receive (but still an issue that we don't exactly know up front if a given node will receive anything at all).  Or could possibly have a separate benchmark for each possible node pairing, but that's a bit cumbersome.

rsmmr:
Nice work. Agree with Jon's questions, it feels like one more layer would be useful that wraps it all together: record all of cluster's activity into a single config, then run a benchmark from that to replay it all back, potentially scaled up. That would allow us to easily run the data collection in a few environments and reproduce their characters on a benchmark box.

Neverlord:
Thanks for the feedback! I should have added a bit more context to the PR. This version is not the "final product". I think of it as an MVP version and the idea with this PR is to get feedback early and iterate a couple of times with follow-up PRs on this. This initial PR also touches a couple of files in Broker to get things working, so I think filing a PR makes sense now to avoid running the branch out-of-sync with `master`.

> > A node cannot send and receive at the same time
> 
> In a real cluster, seems like that's actually a common case to expect

Full ack, this is on my ToDo list. It's simply not implemented yet.

> > At least one node must receive data.
> Seems that's enforced by looking for at least one node with num-inputs ?
> 
> But wonder that a common usage would be "just run the benchmark on the full dataset". In that case, seems a bit awkward to have to explicitly set the expected receive number instead of an option that tells the node "expect to receive unknown amount of data, but upstream publishers will notify you when they're done sending". That way, you still have a termination condition for the node. What do you think?

I would love to get rid of this explicit count. However, there are a couple of factors that make this complicated:
1. The number of publishers per topic is unknown (crucial to count the stop tokens)
2. The benchmark does not know which publisher pushes to which topics
3. Data published to topic A may actually not reach a subscriber for topic A if the Broker endpoints in between do not forward the topic

If I spin your idea of inserting *stop* tokens a bit further:
- We also need to have *start* tokens, then a receiver listens until receiving one *stop* token per previously received *start* token
- A publisher sends the special *start* token before publishing its first data on that topic

Sending the *start* token lazily means to introduce a race. A publisher with very little data can send *start* and *stop* before a second publisher sends its *start*.

We could scan the generator files for all topics at the beginning and then send all *start* tokens eagerly before running the benchmark. However, we still end up with a race since CAF does batching. Publishing the *start* token only puts that message into a buffer at first. If we again have one publisher with only very little data, then it might put its message(s) followed by the *stop* token into the buffer fast enough so that the entire batch reaches a subscriber at once and we have the same problem again.

I think at the very least, we need to compute the number of publishers per topic and subscriber upfront. I think this requires analyzing the distribution tree: get all topics for each publisher and see which subscribers will receive the data (by traveling the tree and stop at in-between nodes that don't forward).

If we do that, we have to scan the generator file anyways. This means we can also count how many messages per topic a publisher generates and from that we can compute how many messages each subscriber gets. So we probably can simply compute `num-inputs` for each node. Scanning the generator files is the expensive part here and might delay each benchmark run for several seconds/minutes (for complex setups). However, once we've computed the `num-inputs` per node we could store the "final" config on disk again. That way, we only pay once for the computation.

> say I have data recorded from a 4 node cluster (a "logger", "worker", "manager", and "proxy"), my first thought was that I should be able to set up a single config file for the benchmark to effectively simulate that cluster environment

Here's the `setup.conf` I used to run the setup with the files you've provided:

```ini
nodes {
  earth {
    id = <tcp://[::1]:8001>
    topics = ["zeek"]
    num-inputs = 303867
  }
  logger {
    id = <local:logger>
    topics = ["zeek"]
    peers = ["earth"]
    generator-file = "zeek-logger-msgs.dat"
  }
  manager {
    id = <local:manager>
    topics = ["zeek"]
    peers = ["earth"]
    generator-file = "zeek-manager-msgs.dat"
  }
  proxy {
    id = <local:proxy>
    topics = ["zeek"]
    peers = ["earth"]
    generator-file = "zeek-proxy-msgs.dat"

  }
  worker {
    id = <local:worker>
    topics = ["zeek"]
    peers = ["earth"]
    generator-file = "zeek-worker-msgs.dat"
  }
}
```

The node `earth` only exists currently because I needed a receiver. Once the nodes can send and receive at the same time this won't be necessary any more.

> have a separate benchmark for each possible node pairing, but that's a bit cumbersome.

Yes, you'll need one config file per peering setup. But I think that's necessary: testing all possible setups automatically explodes very fast. It's less than n^n (since loops are now allowed), but even for four nodes you already have >20 ways to connect them.

> record all of cluster's activity into a single config

I think all we need to auto-generate the `cluster.conf` are peering relations and subscriptions. The `broker-cluster-benchmark` tool would probably not have to change, except for a "generate-config" mode. If we generate the config file than we probably also can do the `num-inputs` calculations.

jsiwek:
> > have a separate benchmark for each possible node pairing, but that's a bit cumbersome.
> 
> Yes, you'll need one config file per peering setup. But I think that's necessary: testing all possible setups automatically explodes very fast.

Yeah, I agree and didn't mean to suggest automatically testing all peering combinations -- I was just pondering workarounds for the "a node cannot both send and receive" limitation, not knowing that's only a temporary issue.  Otherwise, my original impression was a single config file wouldn't accurately define a Zeek cluster.

> > record all of cluster's activity into a single config
> 
> I think all we need to auto-generate the `cluster.conf` are peering relations and subscriptions. The `broker-cluster-benchmark` tool would probably not have to change, except for a "generate-config" mode. If we generate the config file than we probably also can do the `num-inputs` calculations.

Definitely convenient if the correct config file gets auto-generated.

Neverlord:
I've extended the recording capabilities of Broker to generate everything we need to auto-generate the config files in `topic/neverlord/recording`. I also have the auto-generation working so far. However, the generated files now require the benchmark tool to support sending and receiving on the same node.

@jsiwek do you prefer to review/merge the current state first and then do the follow-up PR that adds extra recording etc. or should I extend this PR to include everything?

Neverlord:
Ok, nodes can now enter both modes and I did some cleanup/fixing. 🙂

jsiwek:
Looking great.  Some small touches go a long way like colored error messages and tree diagrams.  

I did notice that a recording from a Zeek cluster seemed to cause the benchmark to hang indefinitely. 
 Can you check this sample recording to diagnose/fix the termination issue ?

[zeek-cluster-recordings.tar.gz](https://github.com/zeek/broker/files/3672737/zeek-cluster-recordings.tar.gz)

I'm using CAF 0.17.1 (our default/embedded CAF in broker/master) if it matters.

Other than getting that representative recording working, I'd say this is ready to merge.

Neverlord:
Thanks for the recording! I've only tested with recordings from `broker-benchmark` so far, which is quite limited TBH. I'll look into it tomorrow.

Neverlord:
I could reproduce the termination issue. However, most of the time it actually crashes for me:

```
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: invalid type found
```

It crashes while trying to deserialize a `caf::variant<caf::cow_tuple<broker::topic, broker::data>, caf::cow_tuple<broker::topic, broker::internal_command>>`. Seems like one of the nodes somehow produces corrupted data. I'm trying to narrow it down.

Neverlord:
@jsiwek sorry, but I'm a bit puzzled. How is that setup working in the first place?

Here's the peering tree I get:

```
logger
└── peers:
    ├── manager
    │   └── peers:
    │       ├── proxy
    │       │   └── peers:
    │       │       └── worker
    │       └── worker
    ├── proxy
    │   └── peers:
    │       └── worker
    └── worker
```

Everyone is connected is to everyone else. In other words: this is a full mesh. My understanding of Broker so far was that we assume a tree structure without loops.

If I change the topology to a tree (choosing `logger` as root here, but anyone would do), I can see all nodes sending their data (but it still won't terminate because the `num-inputs` assumed a different topology):

```
logger
└── peers:
    ├── manager
    ├── proxy
    └── worker
```

Looking at [Manager.cc](https://github.com/zeek/zeek/blob/master/src/broker/Manager.cc), I see code for sending Zeek events [directly to peers](https://github.com/zeek/zeek/blob/master/src/broker/Manager.cc#L488). I didn't scan the entire code base, but I wonder: is Zeek always sending directly? Because otherwise you should see lots of circulating messages that clog the wire until the TTL counter expires when running Zeek.

Either the recording is misleading because it doesn't capture "direct messages", my mental model is wrong, or Zeek uses Broker in an unsafe (and maybe broken) way.

jsiwek:
> Everyone is connected is to everyone else. In other words: this is a full mesh.

Yeah, guess the simplest cluster is, but when multiple logger, worker, or proxy nodes are added to scale things out, those do not connect with each other within the same node type. Example with multiple nodes of a given type:

http://docs.zeek.org/en/stable/frameworks/broker.html#layout-topology

> My understanding of Broker so far was that we assume a tree structure without loops.

Uh, surprise !? :)

* At least the peer graph for a Zeek cluster is not without loops at the moment. 
* The choice of topic names and subscriptions I *think* happens to not cause any message loops at the moment.
* Even if there were certain publication/subscription combos that would send a message into a loop, Zeek disables automatic message forwarding in `broker_options` by default.
* I suppose the use of a TTL in Broker messages is there in the first place as also a way to treat loops, should they occur, so now I wonder where your acyclic assumption was coming from ?  Given our non-use of automatic message forwarding in Zeek, is there still something fundamental broken or just a potential inefficiency in the case someone introduces a message loop via a particular choice of topic/subscription ?

And I guess a short summary of historical reasons why Zeek ended up in this state is:

* The peering topology is close to the what a pre-Broker cluster used.  Not that it has to stay that way, just that was the easiest and most direct first step to take when porting to use Broker.
* A default Zeek cluster didn't seem to have any strong use-case for automatic message forwarding and we also found the current implementation a bit too fragile -- since we do have peering loops, it was too easy to unintentionally introduce message loops.  A future goal related to this was implementing a more sophisticated message routing mechanism to replace the simple forwarding logic that's currently there.  And whatever that is, it would automatically route a message through arbitrary networks to reach all subscribers and also need to handle loops efficiently.

> Looking at [Manager.cc](https://github.com/zeek/zeek/blob/master/src/broker/Manager.cc), I see code for sending Zeek events [directly to peers](https://github.com/zeek/zeek/blob/master/src/broker/Manager.cc#L488). I didn't scan the entire code base, but I wonder: is Zeek always sending directly?

I didn't realize/remember that was there, but no, not always.  That direct send looks like a minority of the messages: one-time upon connecting to a peer, a node directly sends it a list of logs it may write.

Does that still cause problems with the record/benchmark ?  I mean assuming that we don't really care to include them since there's so few of them to make a difference in any benchmarking.

> Because otherwise you should see lots of circulating messages that clog the wire until the TTL counter expires when running Zeek.

Is there really a topic/subscription pairing here that's causing a message to loop ?  Like I said, there was some effort in making sure the default configuration didn't cause any of that, but in the end we just ended up turning off message forwarding by default (so maybe actually new loops did get introduced and just not noticed).

Possibly the recording should write out some of the relevant `broker_options` and mimic that in the benchmark ?

Anything else I can help with for next steps?  I can do another quick recording to see if that one also looks corrupted.  Or if you want to deploy a Zeek cluster yourself to poke around, here's some directions:

* Install Zeek
* Edit `/usr/local/zeek/etc/node.cfg`.  Here's what I used:

```
[logger]
type=logger
host=localhost
env_vars=BROKER_RECORDING_DIRECTORY=/home/jon/tmp/zeek-logger-recording

[manager]
type=manager
host=localhost
env_vars=BROKER_RECORDING_DIRECTORY=/home/jon/tmp/zeek-manager-recording

[proxy-1]
type=proxy
host=localhost
env_vars=BROKER_RECORDING_DIRECTORY=/home/jon/tmp/zeek-proxy-recording

[worker-1]
type=worker
host=localhost
interface=ens33
env_vars=BROKER_RECORDING_DIRECTORY=/home/jon/tmp/zeek-worker-recording
```

You'll have to change the paths in `env_vars` and the `interface` for the worker (also need to run as root or a user with capture permissions for the interface).

* This step probably doesn't matter here, but for completeness if you're just setting up on a local system that probably has checksum offloading enabled, add `ZeekArgs = -C` to `/usr/local/zeek/zeekctl.cfg`.

* Run `zeekctl deploy`

* Generate network traffic on the monitored-interface for some period.

* Run `zeekctl stop`

rsmmr:
Just to add to what Jon's been saying: I think the confusion on topology comes from two different interpretations of loops. Generally, Broker currently cannot deal with routing loops where messages may go in circles. (We have TTLs, but they are meant only as fallback mechanism to catch & tell people there's a problem with their setup; they are not meant as an operational mechanism to deal with loops). However, that doesn't mean that peering cannot set up connections that *could* lead to cycles. Rather, the main thing to watch out for is that the topic subscriptions & forwarding don't actually create loops. We do that (1) by setting topics so that each individually forms just a tree; and (2) also generally not enabling automatic forwarding (with (2) being primarily a safety measure so that people don't shoot themselves into the foot).

rsmmr:
Those tree diagrams are really nice & useful, that's a great debugging tool as well.

Neverlord:
Thank you guys for adding the missing pieces to the puzzle! 🙂

> Zeek disables automatic message forwarding in `broker_options` by default.

This bit in particular was the crucial gap in my mental model as well as in the tool. I've extended the recording mode to also store the full configuration per node to allow `broker-cluster-benchmark` to configure each node exactly as recorded.

I also had to fix a bug in the config generation mode that resulted in miscalculations of `num-inputs`. For better testing, I've also implemented a convenient way to run partial setups by excluding nodes via `-e node1,...,nodeN`.

The tool currently fails to compile in Travis because it needs https://github.com/actor-framework/actor-framework/pull/926. However, CAF `master` already accumulated enough fixes and additions for a new release this week.

> I didn't realize/remember that was there, but no, not always. That direct send looks like a minority of the messages: one-time upon connecting to a peer, a node directly sends it a list of logs it may write.
> 
> Does that still cause problems with the record/benchmark ? I mean assuming that we don't really care to include them since there's so few of them to make a difference in any benchmarking.

If it's only a few than I think we can leave the tool as-is for now, i.e., don't care about direct messages.

> Is there really a topic/subscription pairing here that's causing a message to loop ?

Doesn't look like it. Since the tool sets the `forward` flag accordingly now it wouldn't matter anyway.

> Anything else I can help with for next steps?

You've already helped a great deal just by sharing this information. 👍

I played a bit around with the recordings you've provided (plus adding a `broker.conf` file per directory to set `forward = false`). Generating a config for the full setup and running it works. The benchmark takes about ~6s to complete, but I noticed quite a bit of variance (from 5.3s to 6.2s).

> Those tree diagrams are really nice & useful, that's a great debugging tool as well.

Thanks! They definitely helped me a lot to debug the tool itself. 😉

Neverlord:
FYI, CAF 0.17.2 is out. This PR should build now after updating the submodule.

jsiwek:
Awesome, this was generally working for me now except for #66, but went ahead merging current changes into `master`.

PR #57, opened 2019-09-21
Move config.hh.in out of the include directory

Neverlord:
Having the `config.hh.in` in the `include` directory causes the new [`FindBroker.cmake`](https://github.com/zeek/cmake/pull/9) to erroneously pick up this file instead of looking for the actual `config.hh` in the build tree.


jsiwek:
I had general reservations about https://github.com/zeek/cmake/pull/9 in the first place, but also I'm wondering how/why it can even select `config.hh.in` version if we're explicitly having CMake look for `config.hh` ?  The CMake search logic isn't actually looking for a prefix, is it ?

Maybe in any case, it just seems like the bug would something to solve in `FindBroker.cmake` (add better logic so it doesn't match against the `.in` file) instead of changing the source-tree hierarchy.

Neverlord:
FWIW, I think files that get processed by the build system *are* source files and it semantically makes more sense to place them into an extra `cmake` directory (which is a submodule in Broker, so that's not an option here) or alongside the other sources.

Anyway, see my comment in https://github.com/zeek/cmake/pull/9: it's not actually necessary to move the file.

jsiwek:
> FWIW, I think files that get processed by the build system _are_ source files

Yeah, I can agree w/ that.  Here, I don't feel strongly motivated to make a change anymore, but if you still did, I'd merge it as long as the commit message gets changed to reflect the real reason for moving it

PR #58, opened 2019-09-26
Fix Xcode warnings about subscriber move-assignment

jsiwek:
@Neverlord just a quick sanity check on this when you have time would help

### DIFF
--- a/include/broker/subscriber.hh
+++ b/include/broker/subscriber.hh
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <vector>
+#include <functional>
 
 #include <caf/actor.hpp>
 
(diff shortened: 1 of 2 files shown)


jsiwek:
For reference, warning from latest Xcode looked like:

```
In file included from ../src/broker/Manager.cc:2:
In file included from ../aux/broker/include/broker/broker.hh:15:
../aux/broker/include/broker/subscriber.hh:32:15: warning: explicitly defaulted move assignment operator is implicitly deleted [-Wdefaulted-function-deleted]
  subscriber& operator=(subscriber&&) = default;
              ^
../aux/broker/include/broker/subscriber.hh:58:13: note: move assignment operator of 'subscriber' is implicitly deleted because field 'ep_' is of reference type 'broker::endpoint &'
  endpoint& ep_;
```

PR #67, opened 2019-10-23
Add -DNO_AUTO_LIBCPP when building CAF.

rsmmr:
When setting the C++ compiler to clang, CAF by default switches to
compiling with `-stdlib=libc++`. As Broker doesn't do that, the two end
up linking against different libaries. Setting `NO_AUTO_LIBCPP` lets CAF
use the same library settings as Broker.

(An alternative work-around is telling Zeek/Broker to likewise use libc++ by configuring with `CXXFLAGS="-stdlib=libc++"`. However, aligning defaults seems to be the right fix.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,6 +144,7 @@ else ()
         -DCAF_NO_TOOLS:bool=yes
         -DCAF_NO_PYTHON:bool=yes
         -DCAF_NO_OPENCL:bool=yes
+        -DCAF_NO_AUTO_LIBCPP:bool=yes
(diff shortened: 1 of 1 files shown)


jsiwek:
Helpful to include this patch in Zeek 3.0.1, too?

Neverlord:
> An alternative work-around is telling Zeek/Broker to likewise use libc++

The current patch is the right thing to do. I think CAF should also stop setting such flags by default.

IIRC, the reason we introduced that in CAF was because Clang had problems with some GCC headers. We added that flag for Clang back when GCC 4.7 was the newest version... Time to revisit some of the CMake scaffolding.

Issue #66, opened 2019-10-22
broker-cluster-benchmark sometimes fails to terminate


jsiwek:
Sometimes `broker-cluster-benchmark` doesn't terminate.  E.g. running with the following recordings:

[zeek-broker-recordings.tar.gz](https://github.com/zeek/broker/files/3753458/zeek-broker-recordings.tar.gz)

It sometimes terminates cleanly and other times (maybe every 5-ish times for me) hangs for several minutes so I assume it's not going to terminate.  Here's the verbose output from one of those times:

[broker-cluster-bench-hang.txt](https://github.com/zeek/broker/files/3753460/broker-cluster-bench-hang.txt)

I was able to run the benchmark 40 times without hanging by adding a 2 second `sleep()` after this line:

https://github.com/zeek/broker/blob/b3251b1d6944e398bb97ba824210b895fd886827/tests/benchmark/broker-cluster-benchmark.cc#L1138

So guessing something in the node setup for the benchmark isn't correctly waiting for peerings/subscriptions to be fully established before nodes start sending messages?

Neverlord:
I'm unable to reproduce this with the recordings you've provided. I've tested on my developer box (MacMini, Quad Core) as well as on an AMD Opteron box (CentOS 7, 32 cores). I noticed that the benchmark produces log files and I've [fixed that](https://github.com/zeek/broker/pull/68), but debug and release build both run for me (tested 50x each).

What OS/HW setup are you using? Do you pass any build flags to CAF/Broker other than the build type?

Neverlord:
After running it a bunch more, I've finally seen it in the release build. I'll dig into it some more.

jsiwek:
Thanks for looking into that, combination of #69 and #71 helped me (or at least can now run the benchmark 100 times without hanging where before it was on the order of 10).

PR #68, opened 2019-10-27
Avoid undesired log files when running cluster benchmarks

Neverlord:
While running `broker-cluster-benchmark`, I noticed that it produces log files by default. We're not particularly interested in the logging performance, so this set of changes makes sure the benchmark does not produce log files unless requested by the user.

### DIFF
--- a/tests/benchmark/broker-cluster-benchmark.cc
+++ b/tests/benchmark/broker-cluster-benchmark.cc
@@ -234,6 +234,9 @@ struct node {
 
   /// Stores how many inputs we receive per node.
   inputs_by_node_map inputs_by_node;
+
+  /// Stores the CAF log level for this node.
+  caf::atom_value log_verbosity = caf::atom("quiet");
 };
 
 bool is_sender(const node& x) {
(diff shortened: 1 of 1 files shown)


jsiwek:
Agree that silencing the logs makes sense, but this change seems to introduce an entirely new hanging behavior.  E.g. running same test setup as #66, the benchmark fails to terminate even more frequently.  Don't think platform matters here (tried both macOS and Linux), but can give more precise repro directions if you don't see the same behavior.

I'm running from the build dir like `./bin/broker-cluster-benchmark -c cluster.conf` (or also with the `-v` flag didn't seem to matter).  Here's my `cluster.conf`:

[cluster.conf.txt](https://github.com/zeek/broker/files/3781127/cluster.conf.txt)

Neverlord:
Sorry for not checking this on my end. Hangs for me as well, pretty much at every run. This is also unrelated to #66, rebasing does not help.

Neverlord:
Turned out to be a CAF issue: https://github.com/actor-framework/actor-framework/pull/953.

jsiwek:
Ok, will merge this after CAF 0.17.3 is out then.

Neverlord:
> Ok, will merge this after CAF 0.17.3 is out then.

CAF 0.17.3 [released today](https://github.com/actor-framework/actor-framework/releases/tag/0.17.3).

PR #73, opened 2019-10-30
Set CMAKE_INSTALL_LIBDIR in CAF if set in broker/zeek

0xxon:
With this change, the CAF library files end up in the same directory as the Zeek library files again. Which unbreaks our binary builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,6 +126,12 @@ else ()
     set(cmake_cxx_compiler_launcher_arg)
   endif ()
 
+  if ( CMAKE_INSTALL_LIBDIR )
+    set(cmake_install_libdir_arg -DCMAKE_INSTALL_LIBDIR:path=${CMAKE_INSTALL_LIBDIR})
+  else()
+    set(cmake_install_libdir_arg)
+  endif()
+
   if ( NOT CAF_LOG_LEVEL )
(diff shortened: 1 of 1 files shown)


Neverlord:
Respecting `CMAKE_INSTALL_LIBDIR` seems like the right thing to do.

However, Broker *also* looks for `INSTALL_LIB_DIR` which the `--libdir` option for `configure` sets. @0xxon mentioned a similarly weird behavior in CAF's `--lib-dir`.

@jsiwek do you see a good reason to keep respecting this odd `INSTALL_LIB_DIR` variable? Otherwise, I'd go ahead and open a followup to this PR to set `CMAKE_INSTALL_LIBDIR` in the `configure` script and drop the `INSTALL_LIB_DIR` variable alltogether.

jsiwek:
Good point, seems better to use `CMAKE_INSTALL_LIBDIR` as canonical and remove `INSTALL_LIB_DIR`.  I'll just do that during merge.

One point I can see to keeping `INSTALL_LIB_DIR` (maybe in terms of just redirecting it to set `CMAKE_INSTALL_LIBDIR`) would be if someone has a packaging-related script that uses `cmake -DINSTALL_LIB_DIR` directly or else has Broker as a CMake sub-project and sets that variable.  Don't think those are too likely (or too bad a breakage/change), but open to hear argument from any such people that are following along.

PR #74, opened 2019-11-15
Add workaround to enable parallel Unix make for external project_caf

simonhf:


CMakeLists.txt
```diff
@@ -89,8 +89,15 @@ else ()
     set(use_terminal_arg USES_TERMINAL 1)
   endif ()
 
+  set(caf_make "${CMAKE_MAKE_PROGRAM}")
+  if ( ${CMAKE_GENERATOR} STREQUAL "Unix Makefiles")
```
timwoj: Out of curiosity, does this not happen with `ninja`?

simonhf: The issue does not happen with `ninja`, which appears to have a different architecture from regular Unix make (what cmake calls GNU make). Whereas, without this fix, Unix make builds the external `project_caf` using a single job (concurrently to the regular non-external build which happens in parallel) regardless of the number of jobs specified, my investigations showed that `ninja` appears to use double the amount of jobs for the external `project_caf`, i.e. sort of the opposite of Unix make :-)

jsiwek: Yeah, so since `CMAKE_MAKE_PROGRAM` is `ninja`, it's basically just running `ninja` again with implied parallel build again for CAF.

@simonhf an independent enhancement I just thought of for this patch: should probably check if the generator name ends with "Unix Makefiles" rather than exact equality ?  So then other generators get the benefit, too:

```
$ cmake -G
CMake Error: No generator specified for -G

Generators
* Unix Makefiles               = Generates standard UNIX makefiles.
  Ninja                        = Generates build.ninja files.
  Xcode                        = Generate Xcode project files.
  CodeBlocks - Ninja           = Generates CodeBlocks project files.
  CodeBlocks - Unix Makefiles  = Generates CodeBlocks project files.
  CodeLite - Ninja             = Generates CodeLite project files.
  CodeLite - Unix Makefiles    = Generates CodeLite project files.
  Sublime Text 2 - Ninja       = Generates Sublime Text 2 project files.
  Sublime Text 2 - Unix Makefiles
                               = Generates Sublime Text 2 project files.
  Kate - Ninja                 = Generates Kate project files.
  Kate - Unix Makefiles        = Generates Kate project files.
  Eclipse CDT4 - Ninja         = Generates Eclipse CDT 4.0 project files.
  Eclipse CDT4 - Unix Makefiles= Generates Eclipse CDT 4.0 project files.
```

jsiwek: Nevermind, looks like `CMAKE_GENERATOR` is still set to "Unix Makefiles" for those others, too.

PR #77, opened 2019-11-20
CMake: Add the 'undefined' sanitizer

alessandrogario:
This PR turns the `ENABLE_ADDRESS_SANITIZERS` flag into an option, while also enabling the 'undefined' sanitizer.

This option is now forwarded to the built-in CAF library using `EXTRA_FLAGS` and `CMAKE_EXE_LINKER_SETTINGS`

When the sanitizers are enabled, a number of tests will fail due to either leaks or undefined behaviours.

```
Test project /home/alessandro/Projects/Public/broker/build
      Start  1: backend
 1/17 Test  #1: backend ..........................***Failed    1.09 sec
      Start  2: core
 2/17 Test  #2: core .............................***Failed    0.20 sec
      Start  3: data
 3/17 Test  #3: data .............................   Passed    0.06 sec
      Start  4: data_generator
 4/17 Test  #4: data_generator ...................   Passed    0.06 sec
      Start  5: generator_file_writer
 5/17 Test  #5: generator_file_writer ............   Passed    0.06 sec
      Start  6: meta_command_writer
 6/17 Test  #6: meta_command_writer ..............   Passed    0.06 sec
      Start  7: meta_data_writer
 7/17 Test  #7: meta_data_writer .................   Passed    0.06 sec
      Start  8: integration
 8/17 Test  #8: integration ......................***Failed    0.25 sec
      Start  9: master
 9/17 Test  #9: master ...........................***Failed    0.15 sec
      Start 10: publisher
10/17 Test #10: publisher ........................***Failed    0.12 sec
      Start 11: radix_tree
11/17 Test #11: radix_tree .......................   Passed    4.49 sec
      Start 12: ssl
12/17 Test #12: ssl ..............................***Failed    0.45 sec
      Start 13: status_subscriber
13/17 Test #13: status_subscriber ................***Failed    0.11 sec
      Start 14: store
14/17 Test #14: store ............................***Failed    1.27 sec
      Start 15: subscriber
15/17 Test #15: subscriber .......................***Failed    0.13 sec
      Start 16: topic
16/17 Test #16: topic ............................   Passed    0.06 sec
      Start 17: zeek
17/17 Test #17: zeek .............................   Passed    0.05 sec

47% tests passed, 9 tests failed out of 17

Total Test time (real) =   8.71 sec

The following tests FAILED:
	  1 - backend (Failed)
	  2 - core (Failed)
	  8 - integration (Failed)
	  9 - master (Failed)
	 10 - publisher (Failed)
	 12 - ssl (Failed)
	 13 - status_subscriber (Failed)
	 14 - store (Failed)
	 15 - subscriber (Failed)
Errors while running CTest
```

The full output (with the output of the sanitizers) can be seen by running `ctest -V`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,46 @@ if ( ENABLE_CCACHE )
   set(CMAKE_CXX_COMPILER_LAUNCHER ${CCACHE_PROGRAM})
 endif ()
 
+include(TestBigEndian)
+test_big_endian(BROKER_BIG_ENDIAN)
+
+# Mac OS ignores -pthread but other platforms require it
+if (NOT BROKER_APPLE)
+  set(EXTRA_FLAGS "${EXTRA_FLAGS} -pthread")
+endif ()
+
+# Increase warnings.
+set(EXTRA_FLAGS "${EXTRA_FLAGS} -Wall -Wno-unused -pedantic")
+
+# Increase maximum number of instantiations.
+set(EXTRA_FLAGS "${EXTRA_FLAGS} -ftemplate-depth=512")
+
+option(ENABLE_ADDRESS_SANITIZER "Set to ON to enable the sanitizers")
+
+if ( ENABLE_ADDRESS_SANITIZER )
+  set(additional_flags "-fsanitize=address,undefined -fno-omit-frame-pointer")
+
+  set(EXTRA_FLAGS "${EXTRA_FLAGS} ${additional_flags}")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${additional_flags}")
+  set(CAF_CMAKE_EXE_LINKER_FLAGS "-DCMAKE_EXE_LINKER_FLAGS=${additional_flags}")
+
+  # Some Linux distributions may not always select the correct linker
+  # http://stackoverflow.com/q/37603238/1170277
+  if (BROKER_LINUX)
+    set(EXTRA_FLAGS "${EXTRA_FLAGS} -fuse-ld=gold")
+  endif ()
+
+  if ( NOT "${CMAKE_BUILD_TYPE}" STREQUAL "Debug" )
+    message(WARNING "ENABLE_ADDRESS_SANITIZER requires the build type to be set to Debug")
+  endif ()
+
+  if ( NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" AND
+       NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang" )
+
+    message(WARNING "ENABLE_ADDRESS_SANITIZER requires the compiler to be set to Clang")
+  endif ()
+endif ()
+
 # -- Dependencies -------------------------------------------------------------
 
 if ( CAF_ROOT_DIR )
@@ -163,6 +203,8 @@ else ()
         -DCAF_NO_AUTO_LIBCPP:bool=yes
         -DCAF_BUILD_STATIC:bool=${ENABLE_STATIC}
         -DCAF_BUILD_STATIC_ONLY:bool=${ENABLE_STATIC_ONLY}
+        "-DEXTRA_FLAGS:string=${EXTRA_FLAGS}"
+        ${CAF_CMAKE_EXE_LINKER_FLAGS}
         # Sticking CAF libs in the same output dir as libbroker is a bit of a
         # hack until CMake 3.8, which has BUILD_RPATH that can instead be set
         # on libbroker.  The problem here is that broker unit tests will fail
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -22,6 +22,46 @@ if ( ENABLE_CCACHE )
...
+  endif ()
+
+  if ( NOT "${CMAKE_BUILD_TYPE}" STREQUAL "Debug" )
+    message(WARNING "ENABLE_ADDRESS_SANITIZER requires the build type to be set to Debug")
+  endif ()
+
+  if ( NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" AND
+       NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang" )
+
+    message(WARNING "ENABLE_ADDRESS_SANITIZER requires the compiler to be set to Clang")
```
Neverlord: This is not true. GCC implements the sanitizers as well.

CMakeLists.txt
```diff
@@ -22,6 +22,46 @@ if ( ENABLE_CCACHE )
...
+# Increase warnings.
+set(EXTRA_FLAGS "${EXTRA_FLAGS} -Wall -Wno-unused -pedantic")
+
+# Increase maximum number of instantiations.
+set(EXTRA_FLAGS "${EXTRA_FLAGS} -ftemplate-depth=512")
+
+option(ENABLE_ADDRESS_SANITIZER "Set to ON to enable the sanitizers")
+
+if ( ENABLE_ADDRESS_SANITIZER )
+  set(additional_flags "-fsanitize=address,undefined -fno-omit-frame-pointer")
```
Neverlord: This silently changes the meaning of `ENABLE_ADDRESS_SANITIZER` to a non-obvious behavior. A user that asks for ASAN should not get UBSAN bundled in. Once we add some extra sanitizers by default, where do we draw the line? Some users also might want to enable the thread sanitizer or the memory sanitizer. Eventually, `ENABLE_ADDRESS_SANITIZER` might actually enable all the sanitizers... 😉

I see two approaches work here:
1. Do not touch the `CMakeLists.txt` at all. Users can already set `EXTRA_FLAGS=-fsanitize=address,undefined -fno-omit-frame-pointer` to get the desired outcome without shoe-horning them into exactly ASAN and UBSAN. To make this available to users of the `configure` script, we could instead add a `--extra-flags=` option there. This is the most flexible solution, but certainly requires a bit more "work" on the user side.
2. Drop the `ENABLE_ADDRESS_SANITIZER` flag and instead allow users to list any number of sanitizers (comma separated) in `SANITIZERS`. This is what we do in [`zeek/CMakeLists.txt`](https://github.com/zeek/zeek/blob/master/CMakeLists.txt#L124).

I don't have a strong opinion on which option to pick. @jsiwek do you have a preference between the two?

alessandrogario: Hello @Neverlord!

I have moved the flag handling so that it happens before the built-in CAF is configured (in order to propagate the settings from the parent project to the child one).

The trouble with EXTRA_FLAGS is that we have to set [both the compiler flags and the linker flags](https://clang.llvm.org/docs/AddressSanitizer.html#usage) according to the documentation. It is not possible to pass the flags through that, as they have to land inside the `CMAKE_EXE_LINKER_FLAGS` (i.e. -Wl,-flag would not work).

Ideally, I think the best approach would be to use CMake options with target-based functions (target_link_options/target_compile_options) and avoid touching CMAKE_CXX_FLAGS/CMAKE_EXE_LINKER_FLAGS directly

jsiwek: My thoughts/breakdown:

* For Broker's configuration interface, provide `./configure --sanitizers=` and `BROKER_SANITIZERS` CMake option to users since the `EXTRA_FLAGS` approach might not treat compile versus link flags as precisely we'd need in this case.
* For Broker's implementation of that new flag/option, having it to do the Right thing for both compiler and linker flags sounds fine, but doesn't matter too much to me if it's doing that via `CMAKE_CXX_FLAGS`/`CMAKE_EXE_LINKER_FLAGS` or `target_link_options/target_compile_options`.  The later I guess would be the more Proper/Robust treatment, so feel free to try to get it working via those.
* For the embedded CAF ExternalProject, currently sounds correct to pass sanitizer flags like they are in the PR (`-DEXTRA_FLAGS` and `-DCMAKE_EXE_LINKER_FLAGS`) since `target_compile_options` for `IMPORTED` targets only allows `INTERFACE` items (which I take to mean CAF itself wouldn't be compiled with the desired flags if we went that route)
* For the current meaning of `ENABLE_ADDRESS_SANITIZER`, either don't change it or remove the flag/option entirely since we just provided generalized convenience options for picking sanitizers (I lean more toward the later)

An aside: setting `CXX`/`LD`/`CXXFLAGS`/`LDFLAGS` environment variables is one most-generic way a user can add sanitizer flags (that's originally how I got used to doing it for Zeek), but extending the configure/cmake setup with convenience options does seem helpful.

Neverlord:
> I have moved the flag handling so that it happens before the built-in CAF is configured (in order to propagate the settings from the parent project to the child one).

Ok, that makes sense. Sorry for not catching that in the review.

> For Broker's configuration interface, provide `./configure --sanitizers=` and `BROKER_SANITIZERS`

Thanks for adding your thoughts, @jsiwek. Since the sanitizers need both compiler and linker flags, I think following your suggestion of adding a `BROKER_SANITIZERS` option (and removing `ENABLE_ADDRESS_SANITIZER `) makes the most sense. I forgot about the environment variables, but I agree that adding "convenience features" for very common compile options is worth the while.

> Ideally, I think the best approach would be to use CMake options with target-based functions (target_link_options/target_compile_options) and avoid touching CMAKE_CXX_FLAGS/CMAKE_EXE_LINKER_FLAGS directly

Yes, ideally we wouldn't mess with CMake flag variables directly. We only recently switched to CMake 3 and I think many quirks of the current CMake setup in Zeek/Broker simply have historic reasons. Modernizing the CMake setup should IMO be a separate discussion, though, because it requires a broader discussion. Modernizing the CMake setup could be a followup, but for now I think it makes sense to see how to integrate the sanitizers with minimal changes. 

This is how [Zeek's CMake script](https://github.com/zeek/zeek/blob/master/CMakeLists.txt#L124) sets the sanitizer flags:

```
if ( SANITIZERS )
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=${SANITIZERS} -fno-omit-frame-pointer")
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=${SANITIZERS} -fno-omit-frame-pointer")
    set(CMAKE_LD_FLAGS "${CMAKE_LD_FLAGS} -fsanitize=${SANITIZERS} -fno-omit-frame-pointer")
endif()
```

@alessandrogario do you see any issues with using this template for `BROKER_SANITIZERS` and adding the `configure` option for now?

On a side note, thank you for pushing for more sanitizer usage. I'm already investigating/fixing the UBSan errors for Broker/CAF and once the unit tests under UBSan we should integrate this into our CI builds.

jsiwek:
Since I'm making some general changes to the way Zeek & sub-projects handle sanitizer flags, I'll take care of wrapping this up.  Probably end up just closing the PR since the end solution looks mostly different -- plan on going with the `./configure --sanitizers=LIST` which sets a new `BROKER_SANITIZERS` CMake option, and also passing the necessary flags to CAF external project via `EXTRA_FLAGS`.

The interesting bit I found is that we actually shouldn't need to do anything for the linker flags since CMake defaults to using the compiler to drive linking anyway and automatically picks up the flags that way.  See https://cmake.org/pipermail/cmake/2014-August/058268.html

And if someone is advanced enough to change the default linking behavior of CMake, they'll likely be advanced enough to figure out how to set the flags they require along with it, too, so not inclined to do anything special on our end at least not until we find that user who asks for additional convenience.

PR #78, opened 2019-11-30
Add infinite-send-rate feature to broker-benchmark

Neverlord:
The new mode (enabled via `-r 0` on the command line) makes it easier to measure the maximum throughput Broker can achieve.

### DIFF
--- a/tests/benchmark/broker-benchmark.cc
+++ b/tests/benchmark/broker-benchmark.cc
@@ -229,7 +229,6 @@ void client_mode(endpoint& ep, const std::string& host, int port) {
       // nop
     });
   // Publish events to /benchmark/events.
-  auto p = ep.make_publisher("/benchmark/events");
   // Connect to remote peer.
   if (verbose)
     std::cout << "*** init peering: host = " << host << ", port = " << port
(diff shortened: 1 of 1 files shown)


tests/benchmark/broker-benchmark.cc
```diff
@@ -242,9 +241,30 @@ void client_mode(endpoint& ep, const std::string& host, int port) {
...
+      },
+      [](const caf::unit_t&) { return false; }
+      );
+    for (;;) {
+      // Print status events.
+      auto status_events = ss.poll();
+      if (verbose)
+        for (auto& ev : status_events)
+          std::cout << caf::deep_to_string(ev) << std::endl;
+    }
```
jsiwek: The `ss.poll()` call is non-blocking so this infinite loop spun as fast as it could doing nothing.  Didn't think there's a reason for that so I changed it to `ss.get()` in the merge commit, but let me know if the original way was intentional.

Neverlord: No, that wasn't intentional. I've assumed `poll()` to block for some reason. Thanks for fixing!

PR #79, opened 2020-01-03
Improve sanitizer options/instrumentation

jsiwek:
Part of larger PR at https://github.com/zeek/zeek/pull/722

Also related to #77

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,35 @@ if ( ENABLE_CCACHE )
   set(CMAKE_CXX_COMPILER_LAUNCHER ${CCACHE_PROGRAM})
 endif ()
 
+if ( BROKER_SANITIZERS )
+    # Check the thread library info early as setting compiler flags seems to
+    # interfere with the detection and cause CMAKE_THREAD_LIBS_INIT to not
+    # include -lpthread when it should.
+    find_package(Threads)
+    set(_sanitizer_flags "-fsanitize=${BROKER_SANITIZERS}")
+    set(_sanitizer_flags "${_sanitizer_flags} -fno-omit-frame-pointer")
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -22,6 +22,35 @@ if ( ENABLE_CCACHE )
...
+    if ( NOT DEFINED ENV{NO_OPTIMIZATIONS} )
+      # Using -O1 is generally the suggestion to get more reasonable
+      # performance.  The one downside is it that the compiler may optimize out
+      # code that otherwise generates an error/leak in a -O0 build, but that
+      # should be rare and users mostly will not be running unoptimized builds
+      # in production anyway.
+      set(_sanitizer_flags "${_sanitizer_flags} -O1")
+    endif ()
+
+    # Technically, the we also need to use the compiler to drive linking and
```
timwoj: Minor typo here. There's another a couple lines down.

CMakeLists.txt
```diff
@@ -22,6 +22,35 @@ if ( ENABLE_CCACHE )
...
+    # give the sanitizer flags there, too.  However, CMake, by default, uses
+    # the compiler for linking and so the automatically flags get used.  See
+    # https://cmake.org/pipermail/cmake/2014-August/058268.html
+    set(CAF_EXTRA_FLAGS "${_sanitizer_flags}")
+
+    if ( NOT ZEEK_SANITIZERS )
```
timwoj: Where does this variable come from? Is it only set if we're building broker as part of a zeek build? In that case, is it possible to build broker with address sanitizer outside of that?

timwoj: Scratch that, after looking at the similar commit on the zeek repo I see what's happening here.

timwoj:
Merged with one extra comment about the usage of `ZEEK_SANITIZERS`

Issue #80, opened 2020-01-10
Debugging output not working as expected


rsmmr:
Not sure if I'm missing something, but I'm not getting the debug output I'd expect.

According to an old commit, we have 3 environment vars:

```
        BROKER_DEBUG_LEVEL=<level>
            Set the CAF logging level explicitly.

        BROKER_DEBUG_COMPONENT_FILTER
            Set the CAF component filter explicitly.

        BROKER_DEBUG_VERBOSE=1
            Switch back to old settings: logging at level DEBUG, no
            component filtering.
```

However, no matter how I set these, I don't get any Broker-level messages, but only `caf.flow DEBUG` level messages. Would be good to review how debug logging is working currently and document how to activate it.

Note I noticed this outside of Zeek with the new [zeek-agent](https://github.com/zeek/zeek-agent). That one hardcodes the debug level to -1 currently, which I changed to 2 when trying this. See also https://github.com/zeek/zeek-agent/issues/2

jsiwek:
Running with `BROKER_DEBUG_LEVEL=trace` is what seems to work for me in Zeek/Broker (didn't check into zeek-agent or other log levels).  Examples:

Here's a release Zeek build with no special CAF logging configuration:

```
$ ./configure --build-type=release && make
...
$ BROKER_DEBUG_LEVEL=trace  zeek -r ~/Desktop/get.trace
$ cat broker_78238_1578680818358148000.log
142 caf DEBUG actor0 0x700001db8000 caf.logger log_first_line logger.cpp:613 level = trace, node = 782D1884932E7041B05E3337493D71E11B8BBD00#78238, component-blacklist = []
142 broker INFO actor0 0x11aed2dc0 broker.endpoint endpoint endpoint.cc:195 creating endpoint
142 broker INFO actor0 0x11aed2dc0 ,%20std.vector<topic> subscriber subscriber.cc:156 creating subscriber for topic(s) []
142 broker TRACE actor7 0x700001aa6000 broker operator() core_actor.cc:399 ENTRY filter = []
142 broker TRACE actor7 0x700001aa6000 broker.detail.core_policy add_worker core_policy.cc:365 ENTRY filter = []
...
```

A standalone Broker debug build (don't think debug matters, just what I happened to build) and again without special CAF logging configuration (looks like it defaults to QUIET logging level):

```
$ ./configure --build-type=debug && make install
...
[...]

PR #81, opened 2020-01-18
Redesign Broker's configuration setup

Neverlord:
This set of changes fixes some quirks about Broker's handling of configuration parameters. In particular, it removes the two environment variables `BROKER_DEBUG_VERBOSE` and `BROKER_DEBUG_LEVEL`. They had unclear semantics and did too many (undocumented) things.

The newly recommended way for users to tap into more verbose Broker output is:

- `BROKER_CONSOLE_VERBOSITY=info` for high-level events such as connections/peerings/etc
- `BROKER_CONSOLE_VERBOSITY=debug` for low-level events and seeing what get's published

Users that want the output in a file instead can use `BROKER_FILE_VERBOSITY`.

Please read the detailed commit description of fa952b8 for more technical details. This new set of variables is easy to comprehend and to explain.

Last but not least, I've added a new entry `Troubleshooting` to the documentation that walks users through the available options of getting verbose output. I left out advanced topics such as tweaking the component filter intentionally. Developers that are familiar with CAF can refer to the CAF manual in order to tap into more logging options. Broker users should not have to care about log output of CAF.

I'll also provide a followup to this PR to refine Broker's log output. In particular to make sure users get a better understanding of what went wrong in case of an error. Recent example: Broker's output gives users no clue what went wrong in case two Broker instances with incompatible versions of CAF clash (all they see is a TCP connect followed by a disconnect).

Relates #80.

### DIFF
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -173,6 +173,13 @@ The environment variables take precedence over configuration file entries
 Broker is based on CAF_, so *experienced* users can also use the ``broker.conf``
 to  `tweak various settings
 <https://actor-framework.readthedocs.io/en/stable/ConfiguringActorApplications.html>`_.
+Making use of advanced features is most helpful for developers that contribute
+to Broker's CAF-based C++ source code. For seeing the "full picture", including
+CAF log output, developers can build CAF with log level ``debug`` or ``trace``
+(either by calling ``configure --with-log-level=LVL`` or passing
+``CAF_LOG_LEVEL=LVL`` to CMake directly when using the embedded CAF version) and
+add the entry ``component-blacklist = []`` to the ``logger`` section of the
+``broker.conf`` file.
 
 .. _Zeek: https://www.zeek.org
 .. _CAF: https://actor-framework.org
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -109,4 +109,70 @@ backends, which are currently: in-memory, `SQLite <https://www.sqlite.org>`_, an
 
 :ref:`data-stores` illustrates how to use data stores in different settings.
 
+Troubleshooting
+---------------
+
+By default, Broker keeps console output to a minimum. When running a Broker
+cluster, this bare minimum may omit too much information for troubleshooting.
+
+Users can enable more output either by setting environment variables or by
+providing a ``broker.conf`` file. Custom Broker appliations also may support
+passing command line arguments (Zeek_ does not forward command line arguments to
+Broker).
+
+In order to get a high-level view of what Broker is doing internally, we
+recommend settting:
+
+::
+
+    BROKER_CONSOLE_VERBOSITY=info
+
+Settings this environment variable before running Zeek_ (or any other Broker
+application) prints high-level events such as new network connections, peering
+requests, etc. The runtime cost of enabling this option and the volume of
+printed lines is moderate.
+
+Troubleshooting a Broker application (or Zeek_ scripts that communicate over
+Broker) sometimes requries tapping into the exchanged messages directly. Setting
+the verbosity to debug instead will provide such details:
+
+::
+
+    BROKER_CONSOLE_VERBOSITY=debug
+
+Note that using this verbosity level will slow down Broker and produce a high
+volume of printed output.
+
+Setting ``BROKER_FILE_VERBOSITY`` instead (or in addition) causes Broker to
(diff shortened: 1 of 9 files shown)


Neverlord:
One more note on the implementation: I'm not particularly happy with the current way of propagating errors, i.e., throwing exceptions from constructors. However, using the idiomatic CAF way (calling a function such as `parse` that returns an error) would require changes to Broker's API.

For now, I've opted to keep the API stable but I would like to see the exceptions go eventually. Currently, users can get an exception simply for declaring a variable of type `configuration` (or `endpoint`). This means these Broker types should only live inside a `try` block, which can make the API frustrating to work with and kill your application unexpectedly.

An alternative API could look like this:

```c++
int main(int argc, char** argv) {
  if (auto ep = endpoint::make(argc, argv); // passing (argc, argv) is optional
      not ep) {
    // oops, configuration error
    auto& err = ep.error();
    // ...
  } else {
    // ok, endpoint up and running
  }
}

// With explicitly using a configuration.
int main(int argc, char** argv) {
  configuration conf;
  // ... tweak default settings ...
  if (auto err = conf.init(argc, argv)) {
    // deal with error explicitly, this step is optional and otherwise performed by endoint::run
  }
  if (auto ep = endpoint::make(std::move(conf);
      not ep) {
    // ...
  } else {
    // ...
  }
}
```

In short, I would get rid of all public constructors of `endpoint`. Instead, users would have to go through `endpoint::make`. To an extent, exceptions vs. `error` / `expected` is a matter of style (though I don't think I'm alone when calling constructors that throw particularly bad style). However, a factory function returning an `expected` clearly signals to the programmer that this is an operation that can (and will) fail.

Just my 2¢. 🙂

rsmmr:
Haven't looked at the code yet, but conceptionally this new way of configuring debug output sounds good to me. That will give users an easy way to get what's likely the most useful debug output, and once can still tweak it all further through the configuration file where helpful.

I agree that the `make` factories would be preferable over exceptions, but indeed prefer not to break the API currently. However, we could add them in addition now, and then later deprecate & remove the actors?

Neverlord:
> add them in addition now, and then later deprecate & remove

Sounds like a plan. 👍

I'll do that in a followup.

Neverlord:
> Minor suggestion may be to put a brief sentence in the docs aimed at Broker developers that do want to see CAF debug output.

Done. 🙂

Enabling CAF logs works exactly as you've described: compile CAF with debug/trace logs enabled and reset the filter.

rsmmr:
Nice, merged it. 

@Neverlord A couple notes for when you clean up Broker's debug messages:

- I keep seeing errors about unexpected batch types. I saw that with the zeek-agent already, and now during my testing with `broker-pipe`:

```
# BROKER_CONSOLE_VERBOSITY=info ./broker-pipe -l 12345 -m subscribe
creating endpoint
listening on :12345 (SSL)
creating subscriber for topic(s) []
retrieving connection for 7@68FCC0C79523FF79A6159AFBF655F82238865500#1769 (SSL)
status peer_added ::ffff:127.0.0.1:45748
*** peer_added(68FCC0C79523FF79A6159AFBF655F82238865500#1769, *::ffff:127.0.0.1:45748, "handshake successful")
received unexpected batch type (dropped)
(12345)
```

Any idea what's going in there?

And one request: Can we prefix the debug output with the log level, like `[info]` and `[debug]`, or maybe even `[broker/info]` (not sure what I prefer). That way it becomes clear that it's debug output and why each line is being logged.

Neverlord:
> Any idea what's going in there?

I'll take a look (and track it with a new issue if it's not a trivial fix).

> Can we prefix the debug output with the log level

@rsmmr yes we can! ✨

This is just a configuration option for CAF. Currently, the format string is set to "`%m`", which only prints the log message itself. Here's the full table of supported format string qualifiers:

Screen Shot 2020-01-24 at 12 50 17

Adding the level in square brackets would be simply "`[%p] %m"`". Should I add that to the config or do you think we should add more information while we're at it? I could see a timestamp being useful, for example.

rsmmr:
Nice. Is `%c`going to say "Broker"? Then how about: `[%c/%p] %d %m`?

Neverlord:
> Nice. Is `%c` going to say "Broker"? 

I believe we use lowercase, but yes. LGTM, I'll make a PR.

PR #84, opened 2020-01-24
Fix accidential error/log output

### LINKED ISSUE #83: Invalid / corrupt batches
As reported [here](https://github.com/zeek/broker/pull/81#issuecomment-577691631), `broker-pipe` runs into `received unexpected batch type` errors. Tweaking the log output to print the unexpected batch even results in a segfault.

Neverlord:
Closes #83.

It's one of these bugs where you look at stacktraces for way too long until realizing it's just a missing return. 🤦🏻‍♂️

### DIFF
--- a/src/subscriber.cc
+++ b/src/subscriber.cc
@@ -82,6 +82,7 @@ protected:
       state_->counter += xs_size;
       queue_->produce(xs_size, std::make_move_iterator(xs.begin()),
                       std::make_move_iterator(xs.end()));
+      return;
     }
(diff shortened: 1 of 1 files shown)


jsiwek:
Thanks, I also marked this to be included in Zeek 3.0.2 (via a Broker v1.2.5 patch release) since the misleading debug output seems annoying enough to warrant it.  Or let me know if there's contrary thoughts.

Neverlord:
I think we should include this patch. Really no reason not to.

PR #87, opened 2020-02-02
Support building Broker as CMake subdirectory

Neverlord:
Give parent projects full control over compiler flags by disabling compiler checks and flag management when running as a CMake subdirectory. This set of changes allows the Zeek Agent to build Broker without copying Broker's CMake setup (see https://github.com/zeek/zeek-agent/issues/3).

To achieve "perfect forwarding" of all settings, we also include CAF as subdirectory rather than bundling it as an external project.

Requires the [CAF 0.17.4 branch](https://github.com/zeek/zeek-3rdparty/pull/7), and https://github.com/zeek/cmake/pull/14.


CMakeLists.txt
```diff
@@ -5,320 +5,183 @@ project(broker C CXX)
...
+# leave all compiler flags alone when building as sub directory
+if (NOT broker_is_subproject)
 
-  if ( NOT CCACHE_PROGRAM )
-    message(FATAL_ERROR "ccache not found")
-  endif ()
+  set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bin)
+  set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/lib)
+  set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/lib)
+  set(caf_lib_output_dir ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/${CMAKE_CFG_INTDIR})
```
jsiwek: Seems like `caf_lib_output_dir` is unused now and can be removed.

jsiwek:
Also noticed `CAF_NO_AUTO_LIBCPP` and `CAF_NO_PYTHON` flags don't get passed anymore, but likely should.

And a CMake warning you may be interested in:

```
    CMake Warning (dev) at 3rdparty/caf/CMakeLists.txt:471 (add_custom_command):
      Policy CMP0040 is not set: The target in the TARGET signature of
      add_custom_command() must exist and must be defined in the current
      directory.  Run "cmake --help-policy CMP0040" for policy details.  Use the
      cmake_policy command to set the policy and suppress this warning.

      TARGET 'uninstall' was not created in this directory.
```

Neverlord:
> Also noticed `CAF_NO_AUTO_LIBCPP` and `CAF_NO_PYTHON` flags don't get passed anymore, but likely should.

I've added `CAF_NO_PYTHON`. `CAF_NO_AUTO_LIBCPP` would be redundant, since CAF doesn't touch the flags when building as submodule.

> * CAF headers/libraries are no longer installed via `make install` with this change?  We do rely on those getting installed along with Broker.
> * The `BROKER_APPLE`, `BROKER_LINUX`, and `BROKER_FREEBSD` flags should always get set if we still make them available in `config.hh`.

Fixed. Targets of CAF (including `install`) were suppressed by `EXCLUDE_FROM_ALL`, so I've dropped the flag again. This means CAF builds one or two targets we don't actually need (like the RST generator), but we could fix that later in CAF's CMake setup. Not a big deal at the end of the day, the targets don't take much time to build.

> * Wonder if that setting of the `-pthread` flag is still relevant and something Broker should still take care of itself rather than depend on parent project to set it.

Doesn't the parent has to provide anyways on platforms where it's required? Otherwise, you'd probably run into linker errors anyways.

> If there's an issue with zeek-agent having lack of control over some of the later, maybe I'd add `BROKER_CHECK_SSE2` and `BROKER_CHECK_LIBATOMIC` flags to bypass them, but otherwise seems like the standard usage from Zeek would be right to assume those are automatically done.

Ok, I've moved these two out of the subdir block and guarded each with a new flag for optionally disabling it should the Zeek Agent run into issues with either one.

> And a CMake warning you may be interested in:

Seems like a possible fix could be to add an intermediate custom target in CAF and then add a dependency to `uninstall`. Not sure if that's the way to go, though. For now I'd ignore the warning.

jsiwek:
> > * Wonder if that setting of the `-pthread` flag is still relevant and something Broker should still take care of itself rather than depend on parent project to set it.
> 
> Doesn't the parent has to provide anyways on platforms where it's required? Otherwise, you'd probably run into linker errors anyways.

Good question, not sure, but just noticing that it's currently set regardless of whether Broker is building as submodule and Zeek is building well that way.  Though as long as CI still all builds, would be fine with the change.  I went to check if it did and found it doesn't initially work for at least some other reasons:

https://github.com/zeek/zeek/pull/791

Sorry for not originally trying the build from Zeek myself, but could you help followup to fix those builds?  Some quick observations:

* Possibly CAF includes aren't getting added to search path?
* Looks like `-std=c++17` isn't getting used by Broker when it's included as submodule?

Neverlord:
I'm sorry for not better checking everything is actually working! I thought I've tested with Zeek, but maybe I've ended up testing a build with an external Broker.

> * Possibly CAF includes aren't getting added to search path?

CAF only exported the variable to the parent scope. This means the variables were visible to Broker (the direct parent) but not to Zeek. I've provided a fix in Zeek's CAF fork: https://github.com/zeek/actor-framework/commit/9610cfd1b72ab3bf19ccbd7fc34595fb3c19bf26. I've also fixed the policy warning for the `uninstall` target.

> * Looks like `-std=c++17` isn't getting used by Broker when it's included as submodule?

CMake scoping is weird. But with this patch it seems to work now: https://github.com/zeek/cmake/pull/16

rsmmr:
This would be good to get merged, @jsiwek is this looking good to you?

jsiwek:
> > * Possibly CAF includes aren't getting added to search path?
> 
> CAF only exported the variable to the parent scope. This means the variables were visible to Broker (the direct parent) but not to Zeek. I've provided a fix in Zeek's CAF fork: [zeek/actor-framework@9610cfd](https://github.com/zeek/actor-framework/commit/9610cfd1b72ab3bf19ccbd7fc34595fb3c19bf26). I've also fixed the policy warning for the `uninstall` target.

Would the plan be that those changes can eventually be in upstream CAF ?

> > * Looks like `-std=c++17` isn't getting used by Broker when it's included as submodule?
> 
> CMake scoping is weird. But with this patch it seems to work now: [zeek/cmake#16](https://github.com/zeek/cmake/pull/16)

Now notice a few things probably related to this:

* The Broker/CAF CMake that summarized compiler flags is now misleading when configuring Zeek.  They won't show use of `-std=c++17` because the sub-directories get added before the `RequireCXX17` script that ends up just changing the cache.

*  When configuring Zeek, this warning is now visible for similar reasons to the above (the test was relying on some `-std=`  already being in CMAKE_CXX_FLAGS.

```
-- Performing Test atomic64_builtin
-- Performing Test atomic64_builtin - Failed
-- Performing Test atomic64_with_lib
-- Performing Test atomic64_with_lib - Failed
CMake Warning at aux/broker/CMakeLists.txt:129 (message):
  build may fail due to missing 64-bit atomic support
```

*  Configuring Zeek with `./configure --build-type=debug --sanitizers=address` now fails:

```
CMake Error at cmake/OpenSSLTests.cmake:80 (message):
  OpenSSL library version does not match headers
Call Stack (most recent call first):
  CMakeLists.txt:360 (include)
```

I didn't get a chance to look into explanation for that last one, but now wondering about better ways to to minimize potential unforeseen impact of these changes.  Can we:

* Avoid the new bit about pushing the flags into the CMake cache just to make them globally visible.

* Change the Broker/CAF logic to be like `if ( NOT "${PROJECT_NAME}_NO_COMPILER_FLAGS" )` instead of `if ( NOT "${PROJECT_NAME}_is_subproject" )`.  i.e. just let each project just choose it's own sane/default flags unless the user (or parent project) defines the associated `NO_COMPILER_FLAGS` option and chooses to go their own way.  Then zeek should require no changes and shouldn't see any side effects like we're running into currently, but zeek-agent can start setting that new option.

Neverlord:
> > CAF only exported the variable to the parent scope. This means the variables were visible to Broker (the direct parent) but not to Zeek. I've provided a fix in Zeek's CAF fork: [zeek/actor-framework@9610cfd](https://github.com/zeek/actor-framework/commit/9610cfd1b72ab3bf19ccbd7fc34595fb3c19bf26). I've also fixed the policy warning for the `uninstall` target.
> 
> Would the plan be that those changes can eventually be in upstream CAF ?

Yes. The patches found their way to the [`release/0.17`](https://github.com/actor-framework/actor-framework/tree/release/0.17) branch.

> Now notice a few things probably related to this:
> 
> * The Broker/CAF CMake that summarized compiler flags is now misleading when configuring Zeek.  They won't show use of `-std=c++17` because the sub-directories get added before the `RequireCXX17` script that ends up just changing the cache.
> * When configuring Zeek, this warning is now visible for similar reasons to the above (the test was relying on some `-std=`  already being in CMAKE_CXX_FLAGS.

> * Avoid the new bit about pushing the flags into the CMake cache just to make them globally visible.

I tried to avoid changes to Zeek, but it seems like this was the wrong strategy. The build runs smoothly for me (without any warnings regarding atomics/OpenSSL) with this small patch to Zeek:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index f4401d05c..60e72f872 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,6 +155,8 @@ if ( ZEEK_SANITIZERS )
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${_sanitizer_flags}")
 endif()
 
+include(RequireCXX17)
+
 ########################################################################
 ## Dependency Configuration
 
@@ -360,7 +362,6 @@ include(PCAPTests)
 include(OpenSSLTests)
 include(CheckNameserCompat)
 include(GetArchitecture)
-include(RequireCXX17)
 include(FindKqueue)
 
 if ( (OPENSSL_VERSION VERSION_EQUAL "1.1.0") OR (OPENSSL_VERSION VERSION_GREATER "1.1.0") )
```

Essentially, the patch makes sure we set all flags before entering subdirectories. Setup:

- CAF: `release/0.17` branch (the [Zeek Fork](https://github.com/zeek/actor-framework/tree/release/0.17) is up-to-date for the branch). FYI, the `topic/submodule-build` branch in Zeek's fork should still work, but it seems like you're preferring a bit more "official" version.
- Broker: still `topic/submodule-build`.
- CMake: no longer requires the heavy-handed https://github.com/zeek/cmake/pull/16

> * Change the Broker/CAF logic to be like `if ( NOT "${PROJECT_NAME}_NO_COMPILER_FLAGS" )` instead of `if ( NOT "${PROJECT_NAME}_is_subproject" )`.  i.e. just let each project just choose it's own sane/default flags unless the user (or parent project) defines the associated `NO_COMPILER_FLAGS` option and chooses to go their own way.  Then zeek should require no changes and shouldn't see any side effects like we're running into currently, but zeek-agent can start setting that new option.

With the patch to Zeek's CMake, the subdirectory model now seems to work nicely. I'd like to avoid introducing extra complexity with even more CMake options if possible. I feel like the build setup is already complicated enough as it stands.

Standalone: manage flags ourselves. Subdirectoy: leave flags to the parent. This is a simple enough model to wrap one's head around.

jsiwek:
> > * Change the Broker/CAF logic to be like `if ( NOT "${PROJECT_NAME}_NO_COMPILER_FLAGS" )` instead of `if ( NOT "${PROJECT_NAME}_is_subproject" )`.  i.e. just let each project just choose it's own sane/default flags unless the user (or parent project) defines the associated `NO_COMPILER_FLAGS` option and chooses to go their own way.  Then zeek should require no changes and shouldn't see any side effects like we're running into currently, but zeek-agent can start setting that new option.
> 
> With the patch to Zeek's CMake, the subdirectory model now seems to work nicely. I'd like to avoid introducing extra complexity with even more CMake options if possible. I feel like the build setup is already complicated enough as it stands.
> 
> Standalone: manage flags ourselves. Subdirectoy: leave flags to the parent. This is a simple enough model to wrap one's head around.

I still differ about this: it's maybe simple to summarize in the short-term, but the long-term flexibility and stability seems worse.  It guarantees any change to Broker's minimal flags/requirements needs to be accommodated in multiple places: in standalone broker, in zeek, in zeek-agent, in any other potential projects including it as a submodule.

If zeek-agent is the single instance that ever needs to take full control of the flags, it should maybe be the one explicitly opting into that rather than implicitly causing more potential maintenance burden for any other project that wants to include broker as a submodule.  (I realize the proposed patch to zeek is small and there's likely no other depending projects now, but just going with the "theory" of it).

The complexity in actual CMake script isn't too different: it's still going to be the same conditional branch, just different expression that triggers it.

It also preemptively addresses a generalized use-case: a standalone (non-subproject) build that also wants to take full control of flags.  If that ever came along, we'd just end up having to change/extend the "am I a subproject" logic to check for some flag anyway.

What do you think?

Neverlord:
> > Standalone: manage flags ourselves. Subdirectoy: leave flags to the parent. This is a simple enough model to wrap one's head around.
> 
> If zeek-agent is the single instance that ever needs to take full control of the flags

I see what you're concerned about. And I think we're on the same page here.

By "leave flags to the parent", I meant: base flags. C++ version, standard library (`libc++` vs `libstdc++`), sanitizers (if any), optimization level, and so on. CMake has directory scope for its build flags, so any defines or extra flags Broker sets in addition won't "leak" to the parent. I don't know if we need to introduce a "nuclear option" for the `zeek-agent` where Broker won't even look at the flags the wrong way. Probably not, though.

I think should some flags come up that Broker needs, it should just add them. As long as Broker sets the flags consistently and deterministically, I think it's also fine for the `zeek-agent`. For that reason, I wouldn't add a `BROKER_NO_COMPILER_FLAGS` yet. We'd probably never need it. The crucial thing for `zeek-agent` is that Broker won't pull in any dependency from the system without being told to, because it's building portable binaries.

Should I roll back / bring back some of the setup you deem important in subdirectory mode? My main concern was getting everything to work with `zeek-agent`, so maybe I've erred on the side of cutting things out.

jsiwek:
> I think should some flags come up that Broker needs, it should just add them. As long as Broker sets the flags consistently and deterministically, I think it's also fine for the `zeek-agent`

Yeah, that sounds ok, though I wonder why `-std=XXX` currently doesn't qualify as one of those flags that Broker always needs?

> Should I roll back / bring back some of the setup you deem important in subdirectory mode?

If it works for Broker to always set its own `-std=XXX` (and possibly provide option for overriding what it chooses by default if zeek-agent needs that), that's the only thing being a cause of oddities/confusions at the moment.  I'm game to try merging again once that's clarified.

Neverlord:
Ok, I've moved the C++17-check out of `if`-block. However, I would add one twist to that to make sure Broker does not "downgrade" C++: https://github.com/zeek/cmake/pull/18 (basically, check if C++17 is already available before modifying flags). In a nutshell: if a parent project selects C++20, we currently would downgrade this to C++17. This could result in some unexpected behavior or even break the ABI (whether or not C++ updates may or may not introduce ABI breaks is currently up to debate, but it's definitely better to go with a higher version if the user or parent project explicitly selects it).

PR #90, opened 2020-02-26
Add support for Windows / MSVC

Neverlord:
Notes on porting Broker to MSVC:

- The implementation of `address` relied on some UNIX APIs. Since we're already using CAF, I've simply re-implemented `address` in terms of the platform-neutral `caf::ip_address` (not exposed to the user).
- MSVC is less permissive when it comes to `uniform_int_distribution`. The standard prohibits 8-bit integers and so does MSVC.
- MSVC has the `<filesystem>` header implemented (some GCC releases we still support don't). That made porting the filesystem wrapper mostly straightforward except creating temporary file names (which weirdly enough is not a feature in the C++ standard library).
- Porting socket-related parts was mostly trivial thanks to CAF providing the necessary typedefs etc.
- The lack of `mmap` was a bit tougher to compensate. On Windows, mapping memory is a two-step process.


src/detail/generator_file_reader.cc
```diff
@@ -13,18 +8,105 @@
...
+} // namespace
+
+#else
+
+#include <fcntl.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+std::pair<int, bool> open_file(const char* fname) {
```
jsiwek: Are the non-windows versions of these functions supposed to be in unnamed-namespace like the windows versions are?

src/detail/flare.cc
```diff
@@ -1,69 +1,93 @@
...
+#include <errno.h>
+#include <fcntl.h>
+#include <poll.h>
+#include <unistd.h>
+
+#define PIPE_WRITE ::write
+
+#define PIPE_READ ::read
+
+bool try_again_later() {
```
jsiwek: Expected `try_again_later` and `poll` (for the windows section) to be in unnamed namespace or else `static`.

src/detail/flare.cc
```diff
@@ -76,7 +100,7 @@ size_t flare::extinguish() {
   char tmp[stack_buffer_size];
   size_t result = 0;
   for (;;) {
-    auto n = ::read(fds_[0], tmp, stack_buffer_size);
+    auto n = PIPE_READ(fds_[0], tmp, stack_buffer_size);
     if (n > 0)
       result += static_cast<size_t>(n);
     else if (n == -1 && errno == EAGAIN)
```
jsiwek: This usage of `EAGAIN` didn't change to use `try_again_later()` like other places, that just accidental?

src/detail/generator_file_reader.cc
Neverlord: Yep, good catch.

src/detail/flare.cc
Neverlord: That's an oversight. Fixed.

CMakeLists.txt
```diff
@@ -371,13 +392,15 @@ endif ()
...
-if (NOT BROKER_DISABLE_DOCS)
+if (NOT WIN32 AND NOT BROKER_DISABLE_DOCS)
   add_subdirectory(doc)
 endif ()
 
 # -- Build Summary ------------------------------------------------------------
 
-string(TOUPPER ${CMAKE_BUILD_TYPE} BuildType)
+if (BuildType)
+  string(TOUPPER CMAKE_BUILD_TYPE BuildType)
```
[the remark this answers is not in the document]
Neverlord: To be quite frank, I don't remember what motivated the change. Looking at it again, the change just seems plain wrong. Reverted.

Neverlord:
Thanks for the review! I've included all of your feedback.

> (...) make a PR against Zeek that just updates Broker submodule to point at this branch.
> If you had any specific directions to build/test on Windows, it could help to cover those in the `README`.

Done and done. 🙂

> if Broker is going to officially support Windows, it'll be nice to have that tested in Broker's CI

Seems to be straightforward: https://cirrus-ci.org/guide/windows/. For CAF, we're building the MSVC version via Jenkin's CMake plugin (which ultimately calls `cmake --build ...`). I'm sure Cirrus can do the same.

PR #91, opened 2020-02-26
Send statuses and errors via regular topics

Neverlord:
With this set of changes, status and error messages no longer bypass the regular topic-based distribution. There are two new reserved topics: `topics::statuses` and `topics::errors`. Messages to these two topics are never forwarded to peers.

When emitting status or error messages, they get converted to `data` (consisting of a `vector`). The new conversions between `status <-> data` and `error <-> data` actually accounts for the most code changes. However, this means scripts can now easily access these messages as well. Subscribing to the reserved topics grants easy access to statuses and errors now, allowing to multiplex them with regular events.

The `status_subscriber` was rewritten to dispatch to a regular `subscriber` internally.


jsiwek:
Nicely done.  Minor questions:

* The TODO about replacing the regex involved in `node_id` conversion: seems like that's not too complex to just do now?  Then again, doesn't seem like it's on a critical path (this would only ever happen for status and error messages, right?), so don't mind too much, just slight preference to not leave a TODO around if the intended fix is in easy reach.

* Can you explain the disabling of rate calculation for status subscribers in the unit tests?  Just trying to understand if there's some common pitfall that could generally be avoided.  e.g. if it made sense to always disable rate calculation for status subscriber, because it's better in the general case.  Or if it can just help in the unit testing scenarios then possibly an alternate way to default it to be off otherwise may be easy for people to forget in new unit tests in the future.

And then minor next steps to help boost confidence and get the merge done:

* Travis CI currently reports a build failure that's probably just a trivial missing include

* Just to check that nothing breaks in the Zeek use-case, it can help if another PR is made against Zeek that just has the submodule updated to point to this Broker branch and let it run through its own CI tests: https://github.com/zeek/zeek/pulls

Neverlord:
> The TODO (...) seems like that's not too complex to just do now?

True, but it requires an addition to the CAF API. I've filed PR https://github.com/actor-framework/actor-framework/pull/1058 with the necessary additions. Once that's merged, I'll backport it to `release/0.17` and clean up the TODO.

> Can you explain the disabling of rate calculation for status subscribers in the unit tests?

Sure. 🙂

To calculate the rate, the worker sends itself delayed `tick` messages that arrive every second. In the unit tests, however, this becomes an infinite loop with the deterministic scheduling. If we call `run()` in a unit test, CAF runs all messages until no message remains, then it triggers all timeouts and if any timeouts were triggered, the loop starts over again. In a live system, I don't think the "wakeup message" once per second would cause any trouble. Although one could argue Broker should not run background work until it's asked to by the user. I could get behind a decision to turn it off by default, but I don't have a strong preference here.

> Travis CI currently reports a build failure that's probably just a trivial missing include

After fixing that, it now fails the uint tests on Linux/GCC. I'll look into it.

Once Travis is happy here, I'll file a PR in Zeek.

Neverlord:
> After fixing that, it now fails the uint tests on Linux/GCC. I'll look into it.

I've actually already fixed this. It's a bug in CAF, so we only need to update the submodule (see https://github.com/zeek/zeek-3rdparty/pull/9).

Neverlord:
> And thanks for getting the pending TODO cleanup started (...) comments seemed to stress the "heavy" aspect of it.

It's not on the critical path for Broker, since status and error messages have a low volume. Still better to get this cleaned up right away. Otherwise, the TODO might get forgotten and lurk in there for a long time. Also, the workaround is definitely heavy. `std::regex` isn't cheap (especially if you can replace it with a simple for-loop) and constructing a `node_id` (which does heap allocations) only to throw it away afterwards is quite wasteful.

Backporting to CAF 0.17.x is no big deal, btw. It only means cherry-picking the commits and removing the new export macros.

Neverlord:
@jsiwek I've cleaned up the temporary workaround. Ready to merge on my part. 🙂

PR #97, opened 2020-03-05
Fix a race condition in data store flare operations

jsiwek:
The race (as seen via Zeek usage) goes like:

Thread A: enqueue item, get suspended
Thread B: sees mailbox has items
Thread B: dequeue item
Thread B: extinguish flare
Thread A: resume, fire flare

That ordering can leave the flare in an active state without any actual
items remaining in the mailbox.

This patch adds a mutex/lock such that extinguishing of the flare cannot
be interleaved between the enqueue and firing of the flare.

This likely relates to https://github.com/zeek/zeek/issues/838,
https://github.com/zeek/zeek/issues/716, as well as this thread
http://mailman.icsi.berkeley.edu/pipermail/zeek/2020-February/015062.html

For reference, also attaching a patch that applies against the Broker version shipped with Zeek 3.0.x:

[broker-store-race-3.0.patch.txt](https://github.com/zeek/broker/files/4290901/broker-store-race-3.0.patch.txt)

The CPU load caused by this bug seems particularly exaggerated in Zeek 3.0.x due to some vagaries of its main loop logic: `threading::Manager` is never idle and prevents `FindSoonest` from ever doing the 20 usec sleep when all sources are idle; while the Broker source is also always in the idle state (since it never processes any new messages) but with always-ready FD (due to this bug), meaning the loop tries to process Broker every iteration instead of every 25th iteration had there been an actual way for all sources to be in a legit idle state.

### DIFF
--- a/src/detail/flare_actor.cc
+++ b/src/detail/flare_actor.cc
@@ -27,15 +27,19 @@ void flare_actor::act() {
 
 void flare_actor::await_data() {
   BROKER_DEBUG("awaiting data");
+  std::unique_lock<std::mutex> lock{flare_mtx_};
   if (flare_count_ > 0 )
     return;
+  lock.unlock();
   flare_.await_one();
 }
 
 bool flare_actor::await_data(timeout_type timeout) {
   BROKER_DEBUG("awaiting data with timeout");
+  std::unique_lock<std::mutex> lock{flare_mtx_};
   if (flare_count_ > 0)
     return true;
+  lock.unlock();
   auto res = flare_.await_one(timeout);
   return res;
 }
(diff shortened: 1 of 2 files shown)


src/detail/flare_actor.cc
```diff
@@ -27,22 +27,27 @@ void flare_actor::act() {
 
 void flare_actor::await_data() {
   BROKER_DEBUG("awaiting data");
+  std::unique_lock<std::mutex> lock{flare_mtx_};
```
rsmmr: This would likely work wo/ the lock, but I see you removed the atomic around flare_count and without that, it technically probably isn't guaranteed, so safer to leave in.

rsmmr:
Nice one, looks good to me.

PR #100, opened 2020-03-08
Allow users to subscribe to store updates

Neverlord:
By subscribing to the new (reserved) topic `topics::store_events`, users can now get `insert`, `update`, and `erase` events whenever values in the store are added, changed, or removed.

@0xxon this addition specifically has your use case in Zeek in mind. Let me know if you need additional events from the store. 🙂

The API is rather simple: subscribe to the topic and then you can create views into the `data` like this:

```c++
auto x = subscriber.get();
if (auto insert = store_event::insert::make(x)) {
  // ...
}
// ... same for store_event::update and store_event::erase ...
```


jsiwek:
Just found clarifications for the question of multiple clones attached to same endpoint:

* clone and master stores (using same name) aren't allowed to attach to same endpoint in either Zeek or Broker
* Zeek only allows a single clone of a given store name to attach to a given endpoint (uses a map)
* Broker allows many clones of a given store name to attach to a given endpoint (uses a multimap)

Can double-check me on that last point, but does seem like that may be cause for the larger changes suggested earlier.

include/broker/detail/store_actor.hh
```diff
@@ -0,0 +1,61 @@
...
+
+  /// Emits an `add` event to topics::store_events subscribers.
+  void emit_add_event(const data& key, const data& value,
+                      const caf::optional<timespan>& expiry);
+
+  /// Convenience function for calling
+  /// `emit_add_event(msg.key, msg.value, msg.expiry)`.
+  template <class Message>
+  void emit_add_event(const Message& msg) {
+    emit_add_event(msg.key, msg.value, msg.expiry);
```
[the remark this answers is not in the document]
Neverlord: Yes, I've missed them while renaming. Will fix.

Neverlord:
> * Seems a test is failing w/ the put_unique operation.  One in Broker as seen in CI, and also Zeek has some related failures now.  (Also made the switch to Cirrus CI in `master` recently if you want to bring it in and get wider platform coverage).

I'll look into the failing Python test. If everything's running locally, I'll merge `master` to get the new Cirrus goodness. 👍

> * Pointed out a couple missed renames for the emit functions.

Fixed.

> * For the "update" event, would it be convenient for it to carry both old and new values?  e.g. one reason why it might be important is because by the time one consumes the event, the underlying data already changed and can't actually get at the old value by doing a lookup themselves anymore.

Good point. I've changed to `update` event to include both the new and the old value.

> Now a more significant train of thought [snip]

You've raised some very good points.

> I noticed that these events are all emitted under a single, global topic, `topics::store_events`. I was going to suggest adding other parts to differentiate between stores, like the store name and whether it's a clone or master. So `topics::store_events / store_id / master ? "master" : "clone"`, but then realized there's also no way to differentiate between clones within a single process and you'd possibly get a bunch of "duplicate" events if you had multiple clones? Or do I miss-remember and that's either not possible or a non-use-case ?

I see two things here, so let me try to separate them.

1. *Extending the topic to `topics::store_events / store_id`*: I simply didn't consider the store ID yet. But the keys of course depend on the store and become ambiguous without the store ID. From an API standpoint, there are three ways to encode the store ID: (1) as extra field in the event, (2) as part of the topic as you've suggested, (3) both. Including the store ID in the topic allows subscribers to receive only events they're interested in, but including the ID in the event itself probably makes it easier to process them. So I think I'm leaning towards (3).
2. *Tagging source vs master events*: I can't think of a reason when it should matter to an event consumer whether the event was produced from the master or clone. Do you see a use case where a consumer / script might want to do something with this information?

Related to 2: if we publish the events directly from the master, then subscribers and clones receive the updates separately. As a result, a subscriber can receive an `update_event` before the store receives the `put_command`. This means if a subscriber receives an update and queries its local clone, the clone might still have the old value. The race in the other direction (the store already has a newer value) is unavoidable. Just pointing it out, I can't think of a case where that would hurt from the top of my head.

> Just found clarifications for the question of multiple clones attached to same endpoint

Thanks for digging into this! Honestly, after looking at the code in the `core_actor`, I think we should not spawn multiple clone actors (i.e., use a `map` instead of a `multimap`). All they do is storing the same `std::unordered_map<data, data>`.

There is one reason not to route store subscriptions, though. It replicates data. We cannot implement the clone actor as simple subscriber to the new events, because clones synchronize their state with the master when started. If we always send the `internal_command` to the clone along the events, then we waste bandwidth with events that the clone can produce locally (cheaper). A clone also provides a sort of "history", because it produces `insert_event` messages for all pre-existing values after connecting to a master. Directly subscribing to updates from the master would loose this information.

Of course not routing the store events means one has to attach a clone. This means a subscriber cannot simply listen to *everything*. So in the end it boils down to what use cases we want to support. If we want to allow subscribers to just "get everything" from all stores then we should forward the topic subscriptions. If we want endpoints to explicitly state what store IDs they are interested, then we can safe bandwidth by emitting the updates locally from clones/masters.

So, open questions are:
- Store ID in the topic, in the event itself, or both?
- Optimize for convenience (get-all-subscribers) by routing store topic subscriptions to the master or optimize for bandwidth (require setting up clones locally to receive store events)?

jsiwek:
> 2\. I can't think of a reason when it should matter to an event consumer whether the event was produced from the master or clone. Do you see a use case where a consumer / script might want to do something with this information?

No, my new thinking is it won't be too useful.

> Honestly, after looking at the code in the `core_actor`, I think we should not spawn multiple clone actors (i.e., use a `map` instead of a `multimap`). All they do is storing the same `std::unordered_map<data, data>`.

Yeah, also think we can change it since the use case is narrow -- i.e. may make some unit test code a bit easier, but see no practical use-case.  Historically, may have been more reasons since there was no technical reason to prevent multiple clones and also each clone could configure a different choice of storage backend instead of just in-memory.

> * Store ID in the topic, in the event itself, or both?

Putting it in both seems most helpful/convenient.

> * Optimize for convenience (get-all-subscribers) by routing store topic subscriptions to the master or optimize for bandwidth (require setting up clones locally to receive store events)?

If routing them doesn't also buy us simplified clone implementation (due to synchronization/initialization complications I didn't originally consider), then the wider use-case was kind of speculative on its own anyway.  Think we have more concrete reasons to go the way you originally were without routing them and can also make the `multimap` vs `map` switch to address the ambiguity that originally made me stop and question things.

rsmmr:
@0xxon, how's this Api looking to you?

0xxon:
In general this looks pretty much like what is needed, thank you very much.

One thing that I am not quite sure after looking through this - will this bounce events right back at the sender? So - if a node inserts an element and is subscribed to change notifications, will it get a store event back for that?

If yes - is there any theoretic way to disable that, or to distinguish events from "local" changes from remote changes?

I am mostly trying to think about potential race conditions here.

Neverlord:
> In general this looks pretty much like what is needed

👍

> will it get a store event back for that?

Yes.

> is there any theoretic way to disable that, or to distinguish events from "local" changes from remote changes?

Currently, there's no way to distinguish changes made by the local peer and changes made by other peers. We could include the ID of writing node, but that would only give you node granularity, i.e., you couldn't disambiguate two publishers on the same node. If we need to differentiate this as well, we can include node + publisher ID.

> I am mostly trying to think about potential race conditions here.

If two peers write to their local clone 'simultaneously', then the master performs the writes in whichever order they arrive. This is a race we can't get rid of. We could implement compare-and-swap write operations to at least provide some basic support for atomic operations. Would that help?

rsmmr:
@0xxon ?

0xxon:
I am sorry for the slow reply. I managed to get side-tracked a few times when wanting to answer - and then forgot.

In any case - it would be really neat to be able to distinguish different publishers on the same node. If you could add this - that would be neat.

I am not really concerned about the fact that two peers might end up with a "random" result (based on arrival ordering) - so I don't think we actually need to bother with compare and swap operations.

Thank you very much :)

Neverlord:
Ok. I've introduced a new `publisher_id` type that allows you to keep track of where changes are coming from. You can get this ID from a store by calling `frontend_id()`.

Internally, a publisher ID consists of the endpoint's node ID and an integer. The latter is the actor ID, but for the purpose of Broker, it's just a way to disambiguate store objects on the same endpoint.

Other changes:

- Aside from the synchronization, clones only receive put messages from the master.
- The core actor no longer allows multiple clones for the same topic (`multimap` -> `map` ).

0xxon:
Neat, thanks a lot.

Should we just merge this to master now - or should I try playing around with it in a branch first?

include/broker/detail/master_actor.hh
```diff
@@ -70,21 +68,17 @@ public:
...
   topic clones_topic;
 
   backend_pointer backend;
 
-  caf::actor core;
-
   std::unordered_map<caf::actor_addr, caf::actor> clones;
 
-  endpoint::clock* clock;
+  std::unordered_map<data, publisher_id> publishers;
```
jsiwek, line 77: That's keeping track of who published any given key in the current store?  I wonder if storing that state is even needed.  In the two cases that info gets used, expiration and snapshots, it may make more sense to identify the master store itself as the publisher, not the original/latest publisher that inserted/modified the key?

Looked like that info is used in two case: expiration and snapshots.  For either of those,

jsiwek:
@0xxon before merging, think it's helpful if you get a chance to confirm the approach will work as you expect

0xxon:
That's reasonable - I will try to start working on this.

But - it could take up to few weeks till I have real feedback.

rsmmr:
If it will take a while until we know more, I'd be tempted to just go ahead and merge this in as is, and refine later if we need to. Otherwise we'll have this PR sitting here and just get stale.

jsiwek:
Sure, I'm not strongly against merging earlier and refining if it's going to take a while to play with.  But last thing still wondering about before merge is my latest review comment about the new `publishers` map/state.

include/broker/detail/master_actor.hh
Neverlord: The nice thing about keeping this map is that we can better keep track of who-changed-what last in the subscribers. However, I don't have strong feelings one way or another. Is your concern that his map could become too large?

jsiwek: Yeah, partly thinking the map could become large -- it's replicating the entire key-space.  Other thought is that currently doesn't generalize for use with persistent storage backends.  Could likely do a larger change to the various backends to additionally embed the `publisher_id` along with associated value for a given key to optimize/solve both concerns, but wonder if we really intend to make good use of that info in first place.

Neverlord: Ok, I've removed the `publishers` map from the master.

jsiwek:
All my feedback is addressed (thanks, Dominik).  @0xxon feel free to merge if convenient.

PR #113, opened 2020-05-12
Port Broker to CAF 0.18

Neverlord:
CAF 0.18 brings a new type ID API. Since CAF 0.17.5 includes a subset of the new API, we can support both 0.17.5 and 0.18 with only a few `#ifdef` blocks.

The type ID API assigns ascending IDs to all types at compile time. Types without ID can't be included in messages. A big upside to this is that we no longer have to add individual runtime-type information via `add_message_type`. In 0.17.5, we can pass the entire ID block to `add_message_types`. This means run-time and compile-time type information can no longer run out of sync. CAF 0.18 moves the run-time type information to a global meta object table that Broker needs to fill via `init_global_meta_objects`. This is also a notable change for Zeek, since it has to call `configuration::init_global_meta_objects` once, ideally in its `main`. I'll make the necessary downstream adjustments after merging this PR.

This PR is ready except for updating the CAF reference to 0.17.5 (which is about to be released) and bumping the version check for CAF in CMake.


jsiwek:
Thanks, didn't look closely at code yet since it's draft, but quick comment:

> This is also a notable change for Zeek, since it has to call `configuration::init_global_meta_objects` once, ideally in its `main`. I'll make the necessary downstream adjustments after merging this PR.

Though it sounds like it may be easy even for me to add during/after merge, I generally have slight preference to just see the Zeek-side PR come simultaneously with this if you don't mind.

Neverlord:
No problemo. 🙂

Neverlord:
Thanks for the review. Good points, as usual. :)

I generally try to avoid "implicit magic". But here, I think it definitely makes sense to streamline the experience for Broker users that don't interact with the underlying CAF system. I have used the `call_once` approach you've proposed. I left the static function in the interface in case users want to deploy a custom config class, but I renamed it to `init_global_state`. Making the name a bit more generic makes it the natural place for any static state initialization (just in case we add some in the future).

Adding the deprecated constants is a good idea, I simply didn't think of that. 👍

PR #118, opened 2020-05-28
Add configure argument and install targets for installing benchmarks

timwoj:
This is needed for work on adding broker benchmarking to the CI.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -104,6 +104,8 @@ endif ()
 
 add_executable(broker-benchmark benchmark/broker-benchmark.cc)
 target_link_libraries(broker-benchmark ${libbroker})
+install(TARGETS broker-benchmark DESTINATION bin)
 
 add_executable(broker-cluster-benchmark benchmark/broker-cluster-benchmark.cc)
 target_link_libraries(broker-cluster-benchmark ${libbroker})
+install(TARGETS broker-cluster-benchmark DESTINATION bin)


tests/CMakeLists.txt
```diff
@@ -102,8 +102,12 @@ endif ()
...
 
-add_executable(broker-benchmark benchmark/broker-benchmark.cc)
-target_link_libraries(broker-benchmark ${libbroker})
-
-add_executable(broker-cluster-benchmark benchmark/broker-cluster-benchmark.cc)
-target_link_libraries(broker-cluster-benchmark ${libbroker})
+if (BROKER_ENABLE_BENCHMARKS)
+  add_executable(broker-benchmark benchmark/broker-benchmark.cc)
+  target_link_libraries(broker-benchmark ${libbroker})
+  install(FILES ${CMAKE_BINARY_DIR}/bin/broker-benchmark DESTINATION bin)
```
jsiwek, line 108:
```suggestion
  install(TARGETS broker-benchmark DESTINATION bin)
```

Don't immediately remember if there's substantial difference, but at least a bit shorter this way.

tests/CMakeLists.txt
```diff
@@ -102,8 +102,12 @@ endif ()
...
+
+  add_executable(broker-cluster-benchmark benchmark/broker-cluster-benchmark.cc)
+  target_link_libraries(broker-cluster-benchmark ${libbroker})
+  install(FILES ${CMAKE_BINARY_DIR}/bin/broker-cluster-benchmark DESTINATION bin)
```
jsiwek, line 112:
```suggestion
  install(TARGETS broker-cluster-benchmark DESTINATION bin)
```

timwoj:
> is there a reason to prefer adding the --enable-benchmarks flag?

I don't think there is beyond any reason we weren't already installing them. @Neverlord do you know why they weren't being installed by default before?

Neverlord:
> > is there a reason to prefer adding the --enable-benchmarks flag?
> 
> I don't think there is beyond any reason we weren't already installing them. @Neverlord do you know why they weren't being installed by default before?

No, I can't think of a reason. Probably no one thought of it / needed it.

timwoj:
I backed out the configure changes then and changed the install() commands to how Jon suggested. This should be good to merge now.

PR #121, opened 2020-05-30
Add store ID to insert, update and erase events

### LINKED ISSUE #119: store_events currently give no way to determine the store an event belongs to
This is kind of an addition that was already discussed in https://github.com/zeek/broker/pull/100, but then got forgotten (I think).

It is now possible to receive store events. However there seems to be no way to determine which store an event actually was sent from. Since we will have several of them simultaneously this is kind of important :)

As John noted in #100 , having the store id in the event as well as the topic seems like the best way.

Neverlord:
Closes #119.

Instead of including the store ID in the topic, I have added an additional fields to the event vector. This makes it easier to retrieve the ID and also is less prone to error (there are no restrictions to store IDs, which means the ID can contain a `/` and mess up topic semantics).

### DIFF
--- a/include/broker/store_event.hh
+++ b/include/broker/store_event.hh
@@ -24,6 +24,7 @@ public:
   /// ```
   /// [
   ///   "insert",
+  ///   store_id: string,
   ///   key: data,
   ///   value: data,
   ///   expiry: optional<timespan>,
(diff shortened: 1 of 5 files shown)


jsiwek:
Thanks, looks good, but still some thoughts about including store ID in topic to maybe follow up on:

> I have added an additional fields to the event vector. This makes it easier to retrieve the ID and also is less prone to error

Agree, it's nicer to retrieve store ID from the message rather than parse from topic string, but the use-case for store-id-in-topic may still be for use with subscription filtering: if there's only one store for which a user is interested in events, but there's many other stores that add wasteful overhead of the full message generation/parsing/filtering process, then I think that user might expect that using the existing topic (pre)filtering mechanisms could help minimize the impact.

> (there are no restrictions to store IDs, which means the ID can contain a `/` and mess up topic semantics).

Maybe getting into pedantic side notes, but if such a store ID appears in a topic, I don't see anything messed up about the semantics of topic strings, which is close to "arbitrary sequence of bytes".  The human convention/expectation of `/` being used to delimit and organize hierarchies may be thrown off a bit, but there's still a sensible way to inspect such a topic using prefix-based logic.

0xxon:
Just to add my 2 cents - I agree with Jon here. Having the store ID in the topic too would be neat (and I actually have been using exactly that approach with the patch of Jon for my prototyping).

PR #130, opened 2020-06-29
Include store IDs in the topic

jsiwek:
@Neverlord was just wondering if the PR for this branch got lost and these changes are ok to merge?  Seems like it does what was discussed in #121 and what @0xxon was also thinking would be helpful


Neverlord:
The changes here depended on another PR that got merged already. Seems like it simply fell off my radar to open this PR afterwards. Thanks for filing the PR, one less item on my todo list! 👍

PR #133, opened 2020-07-10
Fix loop detection in the cluster benchmarking tool

### LINKED ISSUE #132: Generating a config file for a recording never terminates
Reported by @timwoj: creating a config file via `broker-cluster-benchmark --generate-config ...` never terminated for a recorded ~30s run of Zeek.

Neverlord:
The `broker-cluster-benchmark` tool did not properly evaluate the `forward` flag. As a result, the loop detection went crazy for setups that looked like loops but weren't due to disabled forwarding. However, the loop detection also relied on a depth heuristic, which could produce false positives.

This PR adds an exact loop detection, prunes redundant peering relations (if A peers to B on startup, we don't need B to peer to A as well), and fixes printing of the peering tree on startup (confused parent/child direction).

Closes #132.

### DIFF
--- a/tests/benchmark/broker-cluster-benchmark.cc
+++ b/tests/benchmark/broker-cluster-benchmark.cc
@@ -226,14 +226,14 @@ struct node {
   bool forward = true;
 
   /// Stores how many messages we produce using the gernerator file. If `none`,
-  // we produce the number of messages in the generator file.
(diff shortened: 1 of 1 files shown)


Neverlord:
@timwoj thanks again for providing the test data. With this patch, I was able to create a configuration file and run the recording locally.

PR #137, opened 2020-07-14
Add integration test for broker-cluster-benchmark

Neverlord:
The new integration test for `broker-cluster-benchmark` uses a (synthetic) recording from @timwoj to check whether both the config generation and the benchmark itself work as expected.

In release mode, the integration test takes ~6 seconds, but a debug build came close to 2min. Hence the bump in `BROKER_TEST_TIMEOUT` to 3min.


jsiwek:
Thanks, minor suggestion: rather than increase `BROKER_TEST_TIMEOUT` globally for all tests, could override for only this one test since none of the others are expected to take anywhere near 2-3min.

(Also in case not already noticed: CI isn't completely happy w/ the new test yet if you can help look into it).

Neverlord:
> Thanks, minor suggestion: rather than increase `BROKER_TEST_TIMEOUT` globally for all tests, could override for only this one test since none of the others are expected to take anywhere near 2-3min.

ACK, I'll override the time only for that test.

> (Also in case not already noticed: CI isn't completely happy w/ the new test yet if you can help look into it).

Looks like all Linux builds run into the timeout. I think I'll need a smaller data set. 😞

Neverlord:
Hm, not sure why Cirrus still runs into timeouts. I've trimmed the recording down to 150 messages in the worker (from >1400). On the CentOS box I'm testing on, the debug build completes in <7 seconds now. A docker image shouldn't slow it down that much, should it?

@jsiwek I'm leaning towards either running the integration tests only in `Release` mode or making them opt-in via CMake/configure option. As long as at least one Cirrus check runs the integration tests, we'll detect breaking changes to the tool(s). What do you think?

jsiwek:
> I'm leaning towards either running the integration tests only in `Release` mode or making them opt-in via CMake/configure option

The current failures in CI are already `Release` builds with no flavor of Linux able to pass that test, so that's suspicious and worth one more step of understanding whether we can reproduce that locally using same Dockerfile (let me know if I can help).  If it does reproduce locally, we should probably figure out the reason.  If it doesn't reproduce locally, then seems fine to settle for skipping that test in Linux CI tasks and possibly make a reminder issue to hunt down what seems like a Cirrus-specific issue.

Neverlord:
> (...) understanding whether we can reproduce that locally using same Dockerfile (...)

TL;DR: Yep. But I couldn't get it working in a Docker container.

Running `debian-9` locally using the `Dockerfile` from Broker:

```
cd ci/debian-9
docker build --tag broker-ci:1.0 .
docker run --rm -it broker-ci:1.0 /bin/bash
```

In the docker image:

```
mkdir /workspace ; cd /workspace
git clone --recursive --branch topic/neverlord/integration-testing https://github.com/zeek/broker.git 
cd broker
./configure --build-type=release --enable-static
make -j 4
cd build
ctest
```

Same behavior. After some digging, I've managed to get to the verbose output of the benchmark:

```
(...)
worker starts listening at [::1]:8002
proxy starts listening at [::1]:8001
logger starts peering to [::1]:8000 (manager)
manager starts listening at [::1]:8000
worker opened port 0 instead of 8002
(...)
```

Seems like the benchmark program is unable to open the configured ports in the Docker image. I've also written the config to disk so that I can run the `broker-cluster-benchmark` tool directly: same result.

The timeout only occurs, because the orderly shutdown doesn't work. I've tweaked the benchmark to call `abort` in case of an error. That way, at least we won't run into timeouts and get hard errors instead.

As to the error itself, seems like the Docker container has trouble with setting up `::1` correctly. Even though `/etc/hosts` looks good:

```
# cat /etc/hosts 
127.0.0.1	localhost
::1	localhost ip6-localhost ip6-loopback
fe00::0	ip6-localnet
ff00::0	ip6-mcastprefix
ff02::1	ip6-allnodes
ff02::2	ip6-allrouters
172.17.0.2	d8ea953c4a11
```

When hardcoding `127.0.0.1` instead, the verbose output looks like opening the ports works. However, the benchmark hangs indefinitely while trying to connect the endpoints:
```
(...)
worker starts listening at 127.0.0.1:8002
logger starts peering to 127.0.0.1:8000 (manager)
worker up and running
proxy starts listening at 127.0.0.1:8001
proxy starts peering to 127.0.0.1:8002 (worker)
manager starts listening at 127.0.0.1:8000
manager starts peering to 127.0.0.1:8001 (proxy)
logger received an error while trying to peer to manager (try again): error(4, 'broker', (invalid-node, *127.0.0.1:8000, "unable to connect to remote peer"))
```

(no further output until killing with CTRL+C)

I don't get why the benchmark has trouble setting up network connections in the container, though. It works in the Python unit tests. They also use `127.0.0.1`.

Any ideas? Running the benchmark on a natively installed CentOS works fine for me, so it's not an issue with Linux.

jsiwek:
I took a look and failed to understand what's going on, too.  Googling the underlying `bind()` error of "cannot assign requested address", I found these Docker issues with similarities:

https://github.com/docker/for-mac/issues/4209
https://github.com/docker/for-mac/issues/4415

Don't seem to apply to the Docker version I used... but makes me doubt that the failure using `::1` is worth hunting down.

However, the other issue when using `127.0.0.1` is more unsettling and has no excuse since we know other Broker/Zeek tests use that addr successfully within Docker.

What was the method used of hardcoding `127.0.0.1` ?  I tried editing the `listen()` and `peer()` called within `broker-cluster-benchmark.cc` as well as within the config file given to `--cluster-config-fie=` and think I got the same behavior as you.  Then as a sanity check, I made the same changes in a working, non-Docker environment and that caused it to stop working, so guess that's not a correct/complete way of hardcoding `127.0.0.1`.

Neverlord:
> (...) but makes me doubt that the failure using `::1` is worth hunting down.

Agree, then let's fall back to the IPv4 localhost address for now.

> What was the method used of hardcoding `127.0.0.1` ?

Replacing the `::1` with `127.0.0.1` during the config generation should be sufficient. However, that alone didn't work as expected due to a bug in CAF. Instead of parsing `127.0.0.1` as an IP address, the URI stores this as a string. When using `to_string` on the variant that stores the host address, the actual string is `"127.0.0.1"`. Due to the quotes, `endpoint::peer` and `endpoint::listen` fail to recognize the input as IP address (or rather, the underlying C functions). With 75786c3, I generate configs with `127.0.0.1` instead of `::1` and provide a custom `to_string` function that makes sure no extra quotes are added. I'll also investigate the CAF side to fix the actual bug and make sure that calling `to_string` on the authority does not add quotes so that we can remove the workaround later (when switching to 0.18).

With these changes, the integration tests run in my local Docker image.

jsiwek:
> When using `to_string` on the variant that stores the host address, the actual string is `"127.0.0.1"`

Thanks for looking into this; the literal quotes sound like such an overlook-able bug.

PR #147, opened 2020-10-21
Update Windows CI config

jsiwek:
* Disable Cirrus Windows task: an updated Docker image on Cirrus-side
  has unknown issues and trying to create one for ourselves also ended
  with similar, unsuccessful results

* Provide an example Dockerfile that uses Visual Studio 2019, which may
  help as a starting point for manual testing/development purposes

* Enable Travis for testing just Windows


jsiwek:
@Neverlord FYI, at least this gets something working again for Windows coverage

PR #149, opened 2020-11-24
Integrate CAF 0.18 into the new ALM backend

Neverlord:
This is a followup to #89 and gets rebased once that hits `master` before this PR can leave *draft* status.

- [x] Require CAF 0.18 via CMake.
- [x] Update submodule version of CAF and migrate to the new CMake options / targets.
- [ ] Remove all `ifdef` statements for supporting CAF < 0.18.


Neverlord:
On a side note: we could also transition to CAF 0.18 when not merging the ALM changes for Zeek 4.0.

That would avoid switching CAF version in Zeek 4.x later, but we should also consider it simply for performance reasons. Broker makes heavy use of CAF streams and they receive some significant performance improvements.

To illustrate, I've used commit 9c3c2a8080cd832ee24f54cdc040cbd0cea91bb2 compiled against CAF 0.17 and CAF 0.18 (this commit is still compatible with both versions). With `broker-benchmark`, I've run a few test runs for all supported event types.

The first run uses event type 1 (a `broker::vector` containing an integer and a string) with 1 million events:

benchmark t1

Just changing the CAF version drops the runtime from 47.75s to 10.88s (the blue line stops earlier, simply because the benchmark is done faster). So just moving to the CAF 0.18 release results in a speedup of >4.

Event type 2 (a `broker::vector` resembling a line in `conn.log`) has lager events and consequently cannot reach the same throughput:

benchmark t2

Still, runtime drops from 50.35s to 17.05s for sending 1 million events. Almost a speedup of 3.

Now, the only case where 0.17 performs slightly better is when using event type 3 (a 100x10 `broker::table`). Note that this time, the benchmark only sends 100,000 messages total.

benchmark t3

The orange line is just a quick sanity check to make sure there's no inherent reason for 0.18 to run slower. By default, CAF assumes small (in terms of bytes) elements in a stream and accordingly has small streaming buffers. I just did one extra run using this `broker.conf`:

```
caf {
  stream {
    size-based-policy {
      buffer-capacity = 524800 # 512KB
      bytes-per-batch =  65536 #  64KB
    }
  }
}
```

That already brings the performance closer. I don't think event type 3 is representative (once we remove the Zeek-side batching), but even if we need to adjust for large messages on average: it just boils down to tuning configuration parameters.

jsiwek:
@Neverlord since CAF 0.18 is merged already from #153, can this get closed as obsoleted or plan to consolidate with #89 ?  Probably would go with #89 as canonical since it contains all more context/discussion.

Neverlord:
Yep, this one became obsolete.

PR #151, opened 2020-11-30
Update Cirrus CI build matrix

Neverlord:
- [x] ~~Re-enable builds via MSVC~~
- [x] Sync up builds with Zeek main repository

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -29,6 +29,14 @@ unix_env: &UNIX_ENV
 # Linux EOL timelines: https://linuxlifecycle.com/
 # Fedora (~13 months): https://fedoraproject.org/wiki/Fedora_Release_Life_Cycle
 
+fedora33_task:
+  container:
+    # Fedora 33 EOL: Around Nov 2021
+    dockerfile: ci/fedora-33/Dockerfile
+    << : *RESOURCES_TEMPLATE
+  << : *CI_TEMPLATE
+  << : *UNIX_ENV
+
 fedora32_task:
   container:
     # Fedora 32 EOL: Around May 2021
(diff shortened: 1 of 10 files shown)


jsiwek:
Think the unresponsive/timeout failure of the Windows CI task is ultimately what I got stuck on last time and ended up moving Windows coverage to Travis CI.  If there's not an issue w/ that, might just want to avoid the time-sink of attempting to move it back to Cirrus.

Neverlord:
>  If there's not an issue w/ that

Not a big issue. Mostly checking whether the windows build is still broken to avoid two separate CI setups. I won't sink much time into that, mostly fiddling while waiting on some Broker/Zeek builds (or benchmark runs).

I do think it'd make sense to keep the build images in sync with the Zeek repo, though. Do you have any plans consolidating those? Maybe the Dockerfiles could go into a Git submodule? That would avoid having two almost identical copies in Broker and Zeek.

jsiwek:
> I do think it'd make sense to keep the build images in sync with the Zeek repo, though.

For sure, at least in terms of platform-coverage, thanks for noticing and looking into it (I had also planned to sync them up this week, but happy to get beaten to it).

> Do you have any plans consolidating those? Maybe the Dockerfiles could go into a Git submodule? That would avoid having two almost identical copies in Broker and Zeek.

I don't have any consolidation plans for now.  Suggest waiting a bit to understand whether periodic syncing of independent Dockerfiles is a big maintenance problem to make consolidation a worthwhile solution.  Besides generally wanting the Dockerfiles used for each repo to express its own "minimal set of dependencies", moving them to common/shared Git submodule still has an awkwardness: verifying any change actually means two PRs to test how it acts in both Zeek and Broker CI contexts (I expect that's not intuitive, easy to forget, or promotes making updates without verification).

Neverlord:
>  Besides generally wanting the Dockerfiles used for each repo to express its own "minimal set of dependencies"

That's a very good point! Didn't think of that. 👍

PR #152, opened 2020-12-02
Upgrade pybind11 to v2.6.1 and fix compatibility issues

jsiwek:
Coming with new Pythons (e.g. 3.9), older pybind11's have some deprecated API usage warnings and also misbehavior; to quote their README:

> Combining older versions of pybind11 (< 2.6.0) with the brand-new Python 3.9.0 will trigger undefined behavior that typically manifests as crashes during interpreter shutdown (but could also destroy your data. You have been warned.)


rsmmr (approved):
Looks good. I had actually just started to look into this as well and was on track to arrive at the same set changes (just hadn't figured out how to add the hashing yet), so easy to review :-)

PR #153, opened 2020-12-04
Migrate to CAF 0.18

Neverlord:
Switch to CAF 0.18 for upcoming Zeek 4.0 release.

PR #155, opened 2020-12-09
Add Python context manager support

jsiwek:
For merging along with #153.  Seems like with these changes, I can get past and no longer reproduce the crash I mentioned in https://github.com/zeek/broker/pull/153#issuecomment-742038602


bindings/python/broker/__init__.py
```diff
@@ -106,13 +106,24 @@ def _make_topics(ts):
     return _broker.VectorTopic(ts)
 
 # This class does not derive from the internal class because we
-# need to pass in existign instances. That means we need to
+# need to pass in existinn instances. That means we need to
```
Neverlord, line 109:
```suggestion
# need to pass in existing instances. That means we need to
```

Neverlord:
Just found that type, otherwise LGTM. 👍

Should I cherry-pick this to the PR branch or are you doing the merging on your end?

jsiwek:
Thanks, I'll just do the merge of this to close out both PRs.

PR #159, opened 2020-12-17
Fix CAF configuration parameter names

Neverlord:
When porting Broker to CAF 0.18, I've forgotten to update the `broker-cluster-benchmark` tool. Our regression test still passed, so I didn't noticed.

Thanks to @timwoj for bringing this to my attention! 🙂


timwoj:
Merged into master and release/2.0.

PR #167, opened 2021-01-07
Implement origin-based credit strategy

### LINKED ISSUE #165: Possible Deadlock in Broker-based Zeek Clusters
In the original design for Broker, we only allowed loop-free topologies. Based on that assumption, we settled on stream-based communication backend to get overload protection via back-pressure. The backend still runs on the assumption of loop-free topologies, but Zeek does not use Broker in this way.

Later additions to Broker introduced the forwarding flag to basically cheat Broker to have loops in the topology while *logically* cutting off all message paths that would go through loops. Zeek does exactly that: it spins up manager, proxy, worker and logger and then forms a full mesh between them while turning forwarding off everywhere. The idea is that the loops exist, but that no message in practice travels more than one hop.

But the loops still exist. The stream-based backend does not know about the forwarding flag and still relies on the assumption that topologies are loop-free.

With recent `master`, the `broker-cluster-benchmark` tool occasionally runs into a deadlock (@timwoj thanks again for providing the recording that reproduced this!). We've migrated to CAF 0.18 recently, so I was worried that it might be CAF bug. After debugging this for a while, I've finally managed to get a trace log and reproduce what's happening. With the benchmarking tool, all Broker endpoints start sending as fast as they can right away. The deadlock manifests itself when:

- the manager waits for credit from the proxy
- the proxy waits for credit from the worker
- the worker waits for credit from the manager

We know that the proxy should give new credit to the manager, because no forwarding to the worker will ever take place. However, the streams don't know that. They have no credit downstream and so they won't grant more credit upstream.

We didn't notice this deadlock before, because CAF 0.17 uses buffers the size of the moon. The oversized buffers made this condition less likely to occur, but once we hit 0 credit on the "wrong" paths, it's game over for Broker. Increasing the buffer sizes again may work as a mitigation strategy, but actually fixing the deadlock won't be straightforward.

I think we still want to keep the back-pressure, but the topic-based messaging logic doesn't fit naturally onto streams any more after allowing loops in the topology. We still want to have streams from local publishers / to local subscribers in order to be able to slow down fast publishers or signal overloads to peers if local subscribers become overloaded. With our loop-free topologies assumption, we simply had one big stream with many paths (to local publishers/subscribers as well as to peers). If we break this up into multiple streams, then we must make sure to include feedback between the streams in order to preserve the back-pressure while making sure to include only the necessary feedback to avoid "false dependencies" that lead to the current deadlock.

Neverlord:
# Intro

Fixes #165 by implementing the "origin-based credit" discussed in the issue. Before going into implementation details, here's a quick sanity check regarding peformance I did with the recordings provided by @timwoj that revealed the deadlock in the first place. I mentioned that the buffer bloat is detrimental to the performance. Mostly due to tons of unnecessary heap allocations and shoving around memory rather than doing actual work. So the natural question is: how bad is it?

This is with the workaround that increases buffer size to 512MB:

```
$ ./build/release/bin/broker-cluster-benchmark --cluster-config-file=cluster.conf
zeek-recording-manager (sending): 1.14407s
zeek-recording-logger (sending): 1.14525s
zeek-recording-proxy (sending): 2.705s
zeek-recording-worker (receiving): 11.6331s
zeek-recording-worker (sending): 12.5222s
zeek-recording-manager (receiving): 28.3079s
zeek-recording-proxy (receiving): 28.9676s
zeek-recording-logger (receiving): 40.3836s
system: 40.383s
```

This is with the new code:

```
$ ./build/release/bin/broker-cluster-benchmark --cluster-config-file=cluster.conf
zeek-recording-logger (sending): 0.000706025s
zeek-recording-manager (sending): 4.13644s
zeek-recording-proxy (sending): 5.79104s
zeek-recording-worker (receiving): 5.80058s
zeek-recording-proxy (receiving): 13.5511s
zeek-recording-manager (receiving): 13.5536s
zeek-recording-worker (sending): 14.0481s
zeek-recording-logger (receiving): 15.5469s
system: 15.5463s
```

YMMV. These numbers are from an AMD Opteron(tm) 6376 1.4GHz with 32 physical cores (64 HW threads).

# Key Concepts

This PR touches a lot of code for a "bugfix". It's fixing a design bug in Broker, though. To read the code, there is a bit of CAF knowledge necessary that I'll cover here first.

## Streaming

Streams provide a back-pressued data channel. Usually connecting two or more actors:

stream

To understand how this maps to actor messages, we also need to take a closer look at message processing. The mailbox of an actor is more than a simple queue. Regular actor-to-actor messages get processed in the order they arrive in according to their priority. But there are two more cases for streaming traffic: downstream messages (for transmitting batches) and upstream messages (for communicating demand). Here's a conceptual view of a mailbox:

mailbox

WDRR stands for Weighted Deficit Round Robin. The takeaway here is that actors multiplex streaming traffic. Streams use `downstream_msg` and `upstream_msg` to allow the actors to put them into the right queues.

## Slots and Paths

Actors can multiplex up to 65k "connections" (max value for an `uint16_t`). Just like TCP uses *ports*, actors use *slots*. Since data only flows in one direction, CAF assigns either an `inbound_path` or an `outbound_path` to each of its slots. These are simple C++ object that keep track of the current state: available/assigned credit, current batch number, batch size, etc.

## Stream Managers

Stream mangers represent a single logical instance of a stream where data arrives on one end (any number of inbound paths) and flows down the other end (any number of outbound paths managed by a `downstream_manager`). In the default model, this is where the backpressure comes in. The stream manager keeps track of how much credit is available downstream and limits credit for the inbound paths based on that.

# How Broker uses Streams Now

The default model does not work for Broker, because we actually have bidirectional communication between peers and our topology includes loops (that we break at a higher level via forwarding flags and TTL). So Broker uses the machinery in CAF for establishing multiplexed message flows with backpressure, but we implement a different credit strategy.

As discussed in #165, we only need to keep track of how many items we currently have in the system per input source (local publisher or peer). The idea is to give each peer X tokens. It may send us that many items and we keep track of the items (each item represents one token). Once we are done with the item (i.e., forwarded it to all subscribers), we can give the peer another token. So the first thing we do with every input is to wrap it into an `item` that tells us where we got this from. To avoid allocating memory for these items over and over again, I've implemented `item_allocator` and `item_stash`. The former allocates blocks of memory for holding some amount of `item` objects. The latter caches up to X items per inbound path. This accomplishes two things: the size of the stash represents the maximum amount of tokens (and what's currently left) and we re-use the same memory every time.

The last piece is the new `unipath_manager`. This is a `caf::stream_manager`, but it only allows at most one inbound path and at most one outbound path. The paths are *not* connected in the usual way, though. The core actor has a single `central_dispatcher` object. All items from all unipath managers get enqueued to the dispatcher. From here, the dispatcher pushes the items to all unicast managers with outbound paths (local subscribers and peers). We hold onto the items until they were shipped to all the right receivers. When the reference count of an item drops to 0, it returns to its stash and we check whether we can give some more credit upstream now.

A stash uses the size `max_pending_inputs_per_source` by default. This is 512 until overriden by a user. This should be plenty. When I played a bit with the recorded data, I usually saw single-digit sized batches and CAF settles at a maximum credit per source ≤ 350. We may need to tweak the default again when we disable the Zeek-side batching.

# Next Steps

So far I've tested the branch using the cluster benchmark and locally with Zeek's btest suites. All changes are internal to Broker, so no API changes. The new way Broker now keeps track of peers makes it easier to keep track of the current handshaking status. As a result, Broker reports errors during handshakes a bit different from before (distinguishes more cases now), which currently causes some Zeek tests for Broker to fail. I'm looking into that to see if we only need to update the baseline. It's still a large diff, so I can understand if you go with the workaround for 4.0 and merge this later. Let me know if this overview is enough to review the changes. When we move forward with this, I'll integrate the new unicast managers into the ALM branch and update the developer guide accordingly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -301,6 +301,7 @@ set(BROKER_SRC
   src/data.cc
   src/defaults.cc
   src/detail/abstract_backend.cc
+  src/detail/central_dispatcher.cc
   src/detail/clone_actor.cc
   src/detail/core_recorder.cc
   src/detail/data_generator.cc
@@ -322,6 +323,7 @@ set(BROKER_SRC
   src/detail/prefix_matcher.cc
   src/detail/sqlite_backend.cc
   src/detail/store_actor.cc
+  src/detail/unipath_manager.cc
   src/endpoint.cc
   src/endpoint_info.cc
   src/error.cc
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -309,6 +309,9 @@ set(BROKER_SRC
   src/detail/flare_actor.cc
   src/detail/generator_file_reader.cc
   src/detail/generator_file_writer.cc
+  src/detail/item.cc
+  src/detail/item_allocator.cc
+  src/detail/item_stash.cc
   src/detail/make_backend.cc
   src/detail/master_actor.cc
   src/detail/master_resolver.cc
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -312,6 +312,7 @@ set(BROKER_SRC
   src/detail/generator_file_writer.cc
   src/detail/item.cc
   src/detail/item_allocator.cc
+  src/detail/item_scope.cc
   src/detail/item_stash.cc
   src/detail/make_backend.cc
   src/detail/master_actor.cc
--- a/include/broker/alm/stream_transport.hh
+++ b/include/broker/alm/stream_transport.hh
@@ -23,9 +23,11 @@
 #include "broker/data.hh"
 #include "broker/defaults.hh"
 #include "broker/detail/assert.hh"
+#include "broker/detail/central_dispatcher.hh"
 #include "broker/detail/filesystem.hh"
 #include "broker/detail/generator_file_writer.hh"
 #include "broker/detail/prefix_matcher.hh"
+#include "broker/detail/unipath_manager.hh"
 #include "broker/error.hh"
 #include "broker/filter_type.hh"
 #include "broker/internal_command.hh"
@@ -40,8 +42,14 @@ namespace broker::alm {
 /// Sets up a configurable stream manager to act as a distribution tree for
 /// Broker.
 template <class Derived, class PeerId>
-class stream_transport : public caf::stream_manager {
+class stream_transport : public detail::unipath_manager::observer {
 public:
+  // -- constants --------------------------------------------------------------
+
+  /// Configures how many (additional) items the stream transport caches for
+  /// published events that bypass streaming.
+  static constexpr size_t item_stash_replenish_size = 32;
+
   // -- member types -----------------------------------------------------------
 
   using peer_id_type = PeerId;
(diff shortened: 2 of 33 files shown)


jsiwek:
Thanks for the helpful descriptions/comments, test-cases, etc.  I pulled this into the release branches for Zeek 4.0.

PR #168, opened 2021-01-17
Clean up maintenance TODOs

Neverlord:
Addresses two `TODO` entries in the master:

- `to_blob` and `from_blob` silently ignore errors (plus use deprecated CAF API)
- we can switch to `caf::hash::fnv` to avoid rolling custom hashing logic


jsiwek (changes requested):
Thanks, can you address the new deprecation warnings that come with this?

Examples: https://api.cirrus-ci.com/v1/task/4674503085129728/logs/build.log

Neverlord:
> Thanks, can you address the new deprecation warnings that come with this?

Huh, interesting. The warnings make sense, but they don't show up for me when building locally. 🤨

I'll take care of them.

rsmmr:
@Neverlord  is this all necessary cleanup right now? Given that we're in RC-mode currently, I'd prefer to keep further changes as minimal as possible (both to reduce churn, but also to lessen review burden)

Neverlord:
> @Neverlord is this all necessary cleanup right now? Given that we're in RC-mode currently, I'd prefer to keep further changes as minimal as possible (both to reduce churn, but also to lessen review burden)

I don't have a strong opinion either way TBH. I filed the PR to fix the TODO regarding silent dropping of errors in `to_blob/from_blob` and to address deprecation warnings, but I hadn't the goal to "make the cut" for the next RC. 🙂

From a user perspective, merging the PR gives a "cleaner build" by addressing the deprecation warnings. Other than that I see this as maintenance that may as well go into the next iteration.

rsmmr:
> > @Neverlord is this all necessary cleanup right now? Given that we're in RC-mode currently, I'd prefer to keep further changes as minimal as possible (both to reduce churn, but also to lessen review burden)
> 
> I don't have a strong opinion either way TBH. I filed the PR to fix the TODO regarding silent dropping of errors in `to_blob/from_blob` and to address deprecation warnings, but I hadn't the goal to "make the cut" for the next RC. 🙂
> 
> From a user perspective, merging the PR gives a "cleaner build" by addressing the deprecation warnings. Other than that I see this as maintenance that may as well go into the next iteration.

Ok, but even just the commit fixing the deprecation warnings seems quite extensive. That's all needed to get them quiet?

Neverlord:
> Ok, but even just the commit fixing the deprecation warnings seems quite extensive. That's all needed to get them quiet?

Most of the code addresses one deprecation: auto-serialization of enums. CAF ≤ 0.17 silently converts between the underlying integer type and the enumeration type. CAF 0.18 now warns about this, because doing this unchecked conversion can break application logic and cause very hard-to-debug errors.

The fix is writing `inspect` overloads for enumeration types. Since I was doing that anyway, I also unified/modernized how Broker deals with conversions to and from enumeration types. CAF 0.18 allows to dispatch on whether an inspector reads/writes binary data or human-readable data such as JSON. So the new code adds `convert` overloads for all enumeration types for the underlying type and strings. This appears much on the diff, but it's mostly just implementing the same API calls for all Broker enums. The new `broker::detail::enum_inspect` then dispatches to the internal conversion API.

So technically this isn't just fixing the warning but also improving the functionality by dispatching on CAF's new `has_human_readable_format()`. I could provide a patch with only the bare minimum (i.e., add `inspect` overloads for all enum types that always dispatch on the underlying type) if you would like to push the added/improved functionality out for the RC.

rsmmr:
Yeah, I think  splitting that out would be good. Then Jon can just cherry-pick that one commit for RC2 if wer don't need anything here else urgently. I'd rather stabilize things now.

Neverlord:
Ok, I've pushed the version dealing with string conversions to a separate branch and pushed a minimal version of the deprecation warning fix here. It's smaller, but feel free to cherry-pick individual commits or force-push into this branch. I got a backup of the original commits. 🙂

PR #171, opened 2021-01-26
Add Travis CI config for Windows OS

jsiwek:


jsiwek:
@Neverlord FYI: this got Windows CI working again via Travis.  Think the way I installed VS 2019 via chocolatey is a bit different than you had, but the main oddity was likely that `-DEXTRA_FLAGS="/MP2"` was passing through CMake shell/quoting and being misinterpreted as a path.

PR #172, opened 2021-02-04
Patch Zeek 4.0 cluster perf/deadlock issues (topic/neverlord/cluster-deadlock-patch)

jsiwek:


jsiwek:
@Neverlord seems the `python-broker-cluster-benchmark` test fails w/ this if you can help any w/ finding out why or let me know if I missed something

Neverlord:
> @Neverlord seems the `python-broker-cluster-benchmark` test fails w/ this if you can help any w/ finding out why or let me know if I missed something

Let me take a look.

Neverlord:
Ok, I've found the bug. Any input from local publishers got dropped because Broker assigned TTL 0 to the messages. It works for Zeek, because the Broker manager in Zeek does not use a local publisher but instead sends publications directly to the core actor. The benchmark uses publishers instead, hence data disappeared.

I've also noticed/fixed a couple minor issues:
- Broker uses the reference count of messages but did not ensure it's 1 at the source. Generally this is the case, but Broker should verify/enforce this condition (fixed by adding/calling the new `force_unshared` functions).
- While benchmarking with @timwoj, I noticed that the unipath managers broke the CAF metrics by not deriving from `downstream_manager_base` and calling appropriate bookkeeping functions (fixed).
- The function `central_dispatcher::ship` added unnecesary overhead since it has O(n). Managers that got new data can simply try to emit batches after enqueue (fixed).
- ASAN reported a heap-buffer-overflow in `broker-benchmark` (fixed).

Giving the branch another run for double-checking performance is probably a good idea.

PR #188, opened 2021-06-18
CI support refresh to bring in line with Zeek

ckreibich:
Drop Fedora 32, add Fedora 34, add Centos Stream, add OpenSUSE Leaps, add FreeBSD 13

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -46,6 +46,14 @@ unix_env: &UNIX_ENV
 # Linux EOL timelines: https://linuxlifecycle.com/
 # Fedora (~13 months): https://fedoraproject.org/wiki/Fedora_Release_Life_Cycle
 
+fedora34_task:
+  container:
+    # Fedora 34 EOL: Around May 2022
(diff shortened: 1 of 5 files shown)


ckreibich:
I'm hitting some instability here with the checks for Centos 8 and also (in earlier runs) other platforms, but at least so far not with the platform's this PR adds. So I'm moving this out of WIP state, but ... does anybody know what's up with the timeout in the Centos 8 benchmark test?

timwoj:
> does anybody know what's up with the timeout in the Centos 8 benchmark test?

Probably the benchmark host having problems again. I'll look into it today.

timwoj:
I fixed the benchmark host. We changed the port number being used by request from Cirrus, and while I updated the variable the zeek repo, I forgot I had to update it here too. You'll need to rebase your branch on master and then it should build cleanly.

timwoj (approved):

ckreibich:
Thanks a bunch Tim, will do.

PR #190, opened 2021-06-25
Fix heap-use-after-free during shutdown

Neverlord:
Fixes https://github.com/zeek/zeek/issues/1607.

@ckreibich with this patch, I can no longer reproduce a crash when hitting CTRL+C. Without the patch, ASAN reported a `heap-use-after-free`. Mind double-checking on your end for the merge?

### DIFF
--- a/src/endpoint.cc
+++ b/src/endpoint.cc
@@ -370,9 +370,8 @@ void endpoint::shutdown() {
     children_.clear();
   }
   BROKER_DEBUG("stop background tasks");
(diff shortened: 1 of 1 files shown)


ckreibich:
Looks good over here, Dominik. Thanks! Both testsuites (Zeek's and Broker's) also pass, fwiw.

Neverlord:
@ckreibich thanks for double-checking! 🙏

PR #197, opened 2021-08-03
Fix publisher deadlock when passing large batches

### LINKED ISSUE #196: Publishing more than #free_capacity events blocks indefinitely
Hi,
we have a Python daemon that publishes events to bro/zeek, which was working fine until we tried to upgrade broker from 1.1.2 to 2.1.0. We noticed that it gets stuck intermittently. After investigating it we realized this happens every time we publish a batch of size `>= publisher.free_capacity()`. 

It was clear that the program was stuck within a native operation: the GIL was taken and the other threads we had in the application were also stuck. The CPU usage was essentially 0 after the application got stuck, indicating that it wasn't caused by getting stuck in an infinite loop.

The issue can be easily reproduced by this short script:

```
import time
import broker


def main():
    ep1 = broker.Endpoint()
    ep2 = broker.Endpoint()
    port = ep1.listen("127.0.0.1", 0)
    subscriber = ep1.make_subscriber("/test")
    ep2.peer("127.0.0.1", port)
    publisher = ep2.make_publisher("/test")
    data = "data"
    publisher.publish_batch(*((data,) for i in range(29)))
    print("29 sent")
    time.sleep(1)
    publisher.publish_batch(*((data,) for i in range(30)))  # blocks indefinitely
    print("this is never reached :(")


if __name__ == '__main__':
    main()

```
The hardcoded capacity of the queue of the publisher is 30,  and apparently whenever it is reached the process gets stuck.
We implemented some logic that makes sure we never publish a batch over the size of the free capacity to avoid this, but it would be nice if this could get fixed upstream.

Bests,
Peter

Neverlord:
Closes #196.

The deadlock is guaranteed when calling `this->fx_.extinguish()` in the shared publisher queue, since we read multiple bytes from the pipe. Without *actually* setting the pipe handle to `nonblock`, this causes the thread to wait forever. I'm a bit surprised that this never bit us before.

### DIFF
--- a/src/detail/flare.cc
+++ b/src/detail/flare.cc
@@ -80,7 +80,7 @@ flare::flare() {
     BROKER_ERROR("failed to set flare fd 0 CLOEXEC: " << res.error());
   if (auto res = child_process_inherit(second, false); !res)
     BROKER_ERROR("failed to set flare fd 1 CLOEXEC: " << res.error());
-  if (auto res = nonblocking(first, false); !res) {
+  if (auto res = nonblocking(first, true); !res) {
     BROKER_ERROR("failed to set flare fd 0 NONBLOCK: " << res.error());
     std::terminate();
   }
--- a/tests/cpp/publisher.cc
+++ b/tests/cpp/publisher.cc
@@ -189,3 +189,37 @@ CAF_TEST(nonblocking_publishers) {
 }
 
 CAF_TEST_FIXTURE_SCOPE_END()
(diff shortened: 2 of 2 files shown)


timwoj:
Merged into master, cherry-picked into release/2.1, updated submodule pointer in zeek 4.1 release branch.

rsmmr:
Oh, good catch (and yeah, surprising)!

PR #201, opened 2021-10-02
Implement new ALM transport

Neverlord:
Picking up from https://github.com/zeek/broker/pull/89 with a major design overhaul to use a custom communication protocol rather than routing on top on CAF's distribution layer.

The new design uses `caf_net` from the [CAF incubator](https://github.com/actor-framework/incubator) as well as a feature branch for the new "CAF flows API", which borrows heavily from ReactiveX. This means this won't work with any official CAF release yet. However, we've already agreed to drop all CAF types from the public API and make the CAF dependency "private" to allow us to have more leeway (and control) of what CAF version we bundle Broker with. Stripping CAF includes from all public Broker headers is a next step after merging this. Mostly mundane work. We only need to move the telemetry wrapper from Zeek one layer up to Broker in order to have Zeek free of any CAF direct CAF dependency.

So where do we stand? Aside from obviously needing test runs. 🙂

I've tested the branch today and it seems like now is a good time to do a feature freeze and review + testing cycle. This represents several weeks worth of new code, so we should expect some new bugs to pop up. I'm going to work on the documentation, but here's the gist of the new design:

- There's a dedicated *connector* in Broker now. This runs in its own thread and performs the handshake. Once a connection is ready to go, it ships the socket over to the core actor.
- After setting up the connection, the core actor starts a socket manager for the new connection. This is a `caf_net` abstraction. All sockets managers run in the *multiplexer* in CAF net. Currently, the multiplexer has a single thread to multiplex all I/O operations.
- The socket manager runs the custom protocol (see `protocol.hh`) and represents a peer. All data it receives from the socket get pushed to a bounded buffer where the core actor picks them up. There's two bounded buffers: one for sending to peers and one for receiving from peers.
- This new bounded buffer is part of the new async API in CAF, which tightly integrates with the new flow API. Inside the core actor, there's basically a branching data flow. Inputs come from local publishers and data stores as well as from remote peers. The core actor combines all inputs to a central data flow (see `central_merge_`). All outputs then filter what they are interested in, e.g., local subscribers only pick data messages targeted to their own peer ID.
- Speaking of publishers and subscribers: I've ripped out the previous shared-queue implementations. Broker consistently uses the new bounded-buffer abstraction and flow integrations wherever it makes sense. The buffers use consumer/producer adapter that allow synchronization such as waking up a consumer on new data. For the consumers and producers, I've implemented adapters to integrate with the flares. This makes the design of Broker much more cohesive by re-using the same abstractions over and over again.

I've had the new version run `broker-benchmark` with a release built for half an hour without issues. Prometheus/Grafana data shows that Broker consumes less than 14MB and hovers slightly above 400% CPU, which is reasonable: 1 core is busy producing data, 1 core is busy receiving/handling the data and one thread per process is busy doing network I/O. The `broker-benchmark` output indicates that Broker averages ~350k msgs/s. The message processing rate shown in Grafana is much lower (about ~30k msgs/s), because actor messages are only uses for demand signaling. The actual data flows through the bounded buffers. So it's probably a good idea to add some custom Broker metrics (after merging this).

I'll wait what bugs surface when running on CI before taking this PR out of draft mode. Zeek's btest suite completed locally with a small fix that avoids running into timeouts by shortening the reconnect interval. So I'll also open up a Zeek-side PR for this to see whether there are still some sporadic timing issues / test failures.


Neverlord:
@rsmmr since your last review, we've added the new networking and switched to the new CAF flow API.

I've drafted this PDF to help navigate the changed architecture and to highlight important abstractions: [2021-10-08 Documentation Draft.pdf](https://github.com/zeek/broker/files/7315037/2021-10-08.Documentation.Draft.pdf). I'm also going to update the official rst files with this content, but I wanted to share this now since I'm AFK for a week and I didn't want you have to wait on the updated Broker doc version before starting to next review iteration.

PR #206, opened 2021-12-14
Improve Python bindings regarding unhashable data structures

### LINKED ISSUE #205: Python bindings fail with mutable index types
Zeek supports mutable index types in containers (for example, a set of records that contain tables), but Python has a rule that index types must be immutable. When Broker encounters such a structure, it crashes. Here's an example:

test.zeek:
```
redef exit_only_after_terminate=T;

type Foo: record {
        config: table[string] of string;
};

global foo: event(s: set[Foo]);

event Broker::peer_added(endpoint: Broker::EndpointInfo, msg: string) {
        local s: set[Foo] = set(Foo($config = table()));
        Broker::publish("foo", foo, s);
}

event zeek_init() {
        Broker::listen();
}
```

test.py:
```
#! /usr/bin/env python

# export PYTHONPATH=$PYTHONPATH:$(zeek-config --python_dir)
import broker

ep = broker.Endpoint()
sub = ep.make_subscriber("foo")

ep.peer("127.0.0.1", 9999)
sub.get()
```

Run `zeek ./test.zeek` in one shell, then `python ./test.py` in another, and you get:
```
$ python /test.py 
Traceback (most recent call last):
  File "/home/christian/t1/./test.py", line 10, in <module>
    sub.get()
  File "/home/christian/inst/opt/zeek/lib64/zeek/python/broker/__init__.py", line 140, in get
    return (msg[0].string(), Data.to_py(msg[1]))
  File "/home/christian/inst/opt/zeek/lib64/zeek/python/broker/__init__.py", line 544, in to_py
    return converters[d.get_type()]()
  File "/home/christian/inst/opt/zeek/lib64/zeek/python/broker/__init__.py", line 540, in <lambda>
    Data.Type.Vector: lambda: to_vector(d.as_vector())
  File "/home/christian/inst/opt/zeek/lib64/zeek/python/broker/__init__.py", line 517, in to_vector
    return tuple(Data.to_py(i) for i in v)
  File "/home/christian/inst/opt/zeek/lib64/zeek/python/broker/__init__.py", line 517, in <genexpr>
    return tuple(Data.to_py(i) for i in v)
  File "/home/christian/inst/opt/zeek/lib64/zeek/python/broker/__init__.py", line 544, in to_py
    return converters[d.get_type()]()
  File "/home/christian/inst/opt/zeek/lib64/zeek/python/broker/__init__.py", line 540, in <lambda>
    Data.Type.Vector: lambda: to_vector(d.as_vector())
  File "/home/christian/inst/opt/zeek/lib64/zeek/python/broker/__init__.py", line 517, in to_vector
    return tuple(Data.to_py(i) for i in v)
  File "/home/christian/inst/opt/zeek/lib64/zeek/python/broker/__init__.py", line 517, in <genexpr>
    return tuple(Data.to_py(i) for i in v)
  File "/home/christian/inst/opt/zeek/lib64/zeek/python/broker/__init__.py", line 544, in to_py
    return converters[d.get_type()]()
  File "/home/christian/inst/opt/zeek/lib64/zeek/python/broker/__init__.py", line 540, in <lambda>
    Data.Type.Vector: lambda: to_vector(d.as_vector())
  File "/home/christian/inst/opt/zeek/lib64/zeek/python/broker/__init__.py", line 517, in to_vector
    return tuple(Data.to_py(i) for i in v)
  File "/home/christian/inst/opt/zeek/lib64/zeek/python/broker/__init__.py", line 517, in <genexpr>
    return tuple(Data.to_py(i) for i in v)
  File "/home/christian/inst/opt/zeek/lib64/zeek/python/broker/__init__.py", line 544, in to_py
    return converters[d.get_type()]()
  File "/home/christian/inst/opt/zeek/lib64/zeek/python/broker/__init__.py", line 534, in <lambda>
    Data.Type.Set: lambda: to_set(d.as_set()),
  File "/home/christian/inst/opt/zeek/lib64/zeek/python/broker/__init__.py", line 511, in to_set
    return set([Data.to_py(i) for i in s])
TypeError: unhashable type: 'dict'
```
This is a well-known problem and the solutions are all variants of making a `dict` derivative hashable. See [here](https://stackoverflow.com/questions/1151658/python-hashable-dicts) for examples. Guido shooting down a `frozendict` PEP is also [worth a read](https://www.python.org/dev/peps/pep-0416/).

So the following is a workaround:

```
-- i/bindings/python/broker/__init__.py
+++ w/bindings/python/broker/__init__.py
@@ -68,6 +68,10 @@
 Configuration = _broker.Configuration
 BrokerOptions = _broker.BrokerOptions

+class HashableDict(dict):
+    def __hash__(self):
+        return hash(frozenset(self))
+
 # Broker's (or better: CAF's) EC code is an integer. Add support
 # for comparision against the enum.
 _EC_eq = _broker.EC.__eq__
@@ -511,7 +515,8 @@
             return set([Data.to_py(i) for i in s])

         def to_table(t):
-            return {Data.to_py(k): Data.to_py(v) for (k, v) in t.items()}
+            tmp = {Data.to_py(k): Data.to_py(v) for (k, v) in t.items()}
+            return HashableDict(tmp.items())

         def to_vector(v):
             return tuple(Data.to_py(i) for i in v)
```

This is clearly incomplete since there can be other combinations of types that would lead to the error. Plus, the object structure returned by `get()` would then have unexpected behaviors in it. One idea is to allow callers to explicitly state that they won't modify the result, perhaps via a separate API or `get(immutable=True)`.

ckreibich:
This is a proposal to addres #205 ... I'm happy to change things if people prefer a different approach, but this branch allows me to move forward with otherwise unrelated cluster controller work (where I first encountered the problem).

Resolves #205

### DIFF
--- a/bindings/python/broker/__init__.py
+++ b/bindings/python/broker/__init__.py
@@ -7,6 +7,7 @@ except ImportError:
 import sys
 import datetime
 import time
+import types
 import ipaddress
 import collections
 
@@ -159,6 +160,36 @@ class Subscriber:
 
     def remove_topic(self, topic, block=False):
         return self._subscriber.remove_topic(_make_topic(topic), block)
+
+class SafeSubscriber(Subscriber):
+    """Subscriber subclass that makes returnes messages safe to process.
+
+    "Safe" here means safe to Python's type system, particularly regarding
+    hashable types. Broker's data model permits nested complex types, such as
+    sets of tables, but those don't directly work in Python (for example,
+    constructing a set of dicts will complain that dicts aren't hashable). To
+    work around this SafeSubscriber relies on ImmutableData rather than Data
+    (used by regular Subscribers)."""
+
+    def get(self, *args, **kwargs):
+        msg = self._subscriber.get(*args, **kwargs)
+
+        if msg is None:
+            return None
+
+        if isinstance(msg, _broker.OptionalSubscriberBaseValueType):
+            if not msg.is_set():
+                return None
+
+            msg = msg.get()
+
+        if isinstance(msg, tuple):
+            return (msg[0].string(), ImmutableData.to_py(msg[1]))
+
+        if isinstance(msg, _broker.VectorPairTopicData):
+            return [(d[0].string(), ImmutableData.to_py(d[1])) for d in msg]
+
+        assert False
 
 class StatusSubscriber():
     def __init__(self, internal_subscriber):
@@ -365,10 +396,17 @@ class Store:
         return (_broker.OptionalTimespan(_broker.Timespan(float(e))) if e is not None else _broker.OptionalTimespan())
 
 class Endpoint(_broker.Endpoint):
-    def make_subscriber(self, topics, qsize = 20):
+    def make_subscriber(self, topics, qsize = 20, subscriber_class=Subscriber):
         topics = _make_topics(topics)
         s = _broker.Endpoint.make_subscriber(self, topics, qsize)
-        return Subscriber(s)
+        return subscriber_class(s)
+
+    def make_safe_subscriber(self, topics, qsize = 20):
+        """A variant of make_subscriber that returns a SafeSubscriber instance. In
+        contrast to the Subscriber class, messages retrieved from
+        SafeSubscribers use immutable, hashable values to ensure Python can
+        represent them. When in doubt, use make_safe_subscriber()."""
+        return self.make_subscriber(topics=topics, qsize=qsize, subscriber_class=SafeSubscriber)
 
     def make_status_subscriber(self, receive_statuses=False):
         s = _broker.Endpoint.make_status_subscriber(self, receive_statuses)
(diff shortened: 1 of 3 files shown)


Neverlord:
Just wondering, if `Subscriber` is broken is there any benefit of leaving it available to users rather than just change the `Subscriber` class instead of providing a "safe" alternative?

Neverlord:
Makes sense. I guess the question is whether we consider it a bugfix (that happens to requires an API change) or more of a long/mid-term transition to safer APIs.

> perhaps your ALM work implies larger Broker changes?

Not on the Python side of things. But the internals change a lot, so once we merge that I think a major version bump is necessary.

ckreibich:
Got it. So the course of action would be to add `SafeSubscriber` now for Broker 2.2 / Zeek 4.2 and deprecate it with Broker 3 / Zeek 5 (assuming ALM in Zeek 5), when its behavior would become the default. Does that sound good? I can technically work around the problem for now, but we'd leave the problem unfixed for a while ... so I'd rather not.

rsmmr (approved):
This is a nice solution.

PR #207, opened 2021-12-17
Avoid exposing CAF internals

### LINKED ISSUE #144: Clean up Broker API to avoid exposing internals
(Moved here from Zeek issue tracker.)

The Broker C++ API still exposes some CAF internals. We should clean that up to become standalone. That could just mean some using X = caf::X but in the end, API consumers shouldn't be referring to CAF directly anymore.

Neverlord:
*NOTE*: before we merge this, we'll first need to add a new wrapper for the CAF metrics that Zeek can use. Currently, Zeek relies on accessing CAF directly which is no longer possible after merging this. That's going to be straightforward and I'll start working on it right away.

This is a big diff, but almost all changes are trivial. Changed header paths, using `std::` instead of `caf::`, things like that. 🙂

New code only provides a way to make CAF handles opaque (like `broker::worker`) or copies CAF types into the Broker codebase (plus light adjustments) like `broker::cow_tuple`.

The goal is to remove all CAF dependencies from public headers and to build the bundled CAF version as static, private utility. Consumers of Broker no longer need to locate CAF headers and we do not install any CAF content alongside Broker, i.e., Broker becomes fully standalone.

Where possible, Broker uses recent C++ additions like `std::variant` to replace CAF types. For types that have no equivalent, Broker deploys fully opaque handle types. For example, `broker::worker` internally wraps a `caf::actor` handle.

All components in Broker that require access to CAF were grouped into the new namespace `broker::internal`. We omit all internal headers from an installation, since they would be worthless without the matching CAF headers anyways. This also gives us a clean separation between API artifacts users may interact with and API parts that are reserved for internal use.

Closes #144.


Neverlord:
The [CI error on OpenSuse Leap 15.2](https://github.com/zeek/broker/pull/207/checks?check_run_id=4568958095) seems unrelated. This PR isn't touching any functionality and we've seen this test time out before on master. For example [here](https://github.com/zeek/broker/runs/4121267738) and [here](https://github.com/zeek/broker/runs/2972127120).

rsmmr (commented):
Nice, thanks, I briefly looked over most of the changes, and that looked all good to me. Most changes are indeed straight-forward/expected, and I didn't notice anything particular otherwise.

That said, I'm leaning towards holding off merging until we cut 4.2 (even though it'd be nice to get this in earlier). Couple reasons:

1. A few pieces are necessarily changing namespaces as some of the `caf::..` IDs were part of the API. Most won't be affecting anybody, but we did show some of those in out docs, so presumably there are people out there using them. We could try to maintain aliases and add deprecation warnings, but not sure that's worth the effort. But without that, pulling this in early in a cycle seems better that very late.

2. The changes are a bit more involved than I was hoping originally. It all makes sense, but there are a few chunks of new code in there, providing replacements for previous functionality. Some of that is directly taken from CAF, but there's new Broker-only code as well, so there'd be some risk in merging this just before 4.2, even if low.

How do other's feel?

rsmmr:
Is it possible that this breaks the handshake against current Broker versions? Using a custom Broker client app, when I switch from Broker `master` to this branch, the handshake for the initial peering doesn't seem to finish anymore. Communication just hangs and when I shut down Zeek, the client reports `failed to complete handshake`. Not sure if this is expected, but it would be good to avoid.

Neverlord:
> Is it possible that this breaks the handshake against current Broker versions?

It shouldn't. Both the `master` branch and the `issue/144` point to the same CAF version (0.18.5). Just to be sure, did you maybe build one version against a locally installed CAF version and the other using the builtin version? Is there some additional Broker log output to the error? The setup for setting the CAF application ID also remained the same: https://github.com/zeek/broker/blob/issue/144/src/configuration.cc#L138.

Do you have the broker versions still on disk? One thing I'd try is whether the `broker-benchmark` tool works "cross-version", i.e., running `broker-benchmark` from `master` as server and then connecting `broker-benchmark` from `issue/144` as client to it. If you no longer have the artifact I'll try reproducing it this way.

rsmmr:
> One thing I'd try is whether the `broker-benchmark` tool works "cross-version"

Good idea. I tried that, and that does reproduce the problem. I re-compiled Broker twice, once from master, once from this branch. Same configure call with `--enable-debug --enable-static-only`, just with the branched switched in between.

I'm attaching two client-side logs from `BROKER_FILE_VERBOSITY=trace`, one for each run, both against running a `master` server
[broker_issue_144.txt](https://github.com/zeek/broker/files/7841090/broker_issue_144.txt)
[broker_master.txt](https://github.com/zeek/broker/files/7841091/broker_master.txt)
.

rsmmr (commented):
Couple more things I noticed:

- the tests still use a lot of `caf::*`.  I imagine that's not completely avoidable, but would be good to switch over wherever we can, because tests also show people how to use the API.

- If somebody installs this over an existing Broker version, the headers will get messy because of the `internal` vs `detail` change. How about fully removing the target `detail` directory before installing the new headers?

Otherwise, looks good to me (but hard to review in detail).

Neverlord:
> the tests still use a lot of `caf::*`. I imagine that's not completely avoidable

Some components, like the core actor, are actors. So testing them naturally requires access to CAF. However, I did a pass over the unit test suites and some places really ought to use the Broker API level instead. In particular the publisher and subscriber unit tests. I've rewritten those tests with Broker-native constructs. With some help from the fixture to take care of deterministic execution and peering (without actually doing network).

> How about fully removing the target detail directory before installing the new headers?

Done. I've also [fixed a bug](https://github.com/zeek/broker/pull/207/commits/e3552b86bba3f8832793548aa0a119ad68911468#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aL241) in the new CMake setup that I didn't notice until now.

rsmmr:
This works for me now, thanks!

One more question: I don't see a replacement for `caf::unit_t`, which some of the callback needs in their signature. In the unit tests, I see `caf::unit_t` still being used, whereas `broker-benchmark` defines its own `struct no_state {};`. Should we make the latter available inside the Broker namespace as either `broker::no_state` or `broker::unit_t`?

rsmmr:
@neverlord Any thoughts on the above? Other than that, I think this is ready to merge and I'd go ahead with that then.

Neverlord:
@rsmmr yes, sorry for the delay. I think adding `no_state` as public type makes sense. We could also make the state parameter optional to make this hacky construct obsolete to begin with. I tend toward the latter. What would you say?

rsmmr:
> @rsmmr yes, sorry for the delay. I think adding `no_state` as public type makes sense. We could also make the state parameter optional to make this hacky construct obsolete to begin with. I tend toward the latter. What would you say?

Yes, that'd be great, I've been wondering about whether we need that. Can you make that change then?

Neverlord:
Will do.

Issue #213, opened 2022-02-25
Build is broken on Windows


timwoj:
I'm trying to build with VS 2022. It builds fine from the 2.2.0 tag, but fails on `master`.

Neverlord:
This should have been flagged by Travis, but apparently Windows builds (silently?) stopped a while ago.

timwoj:
I started looking into getting our Cirrus Windows builds going again on
Friday but I ran out of time to finish it. I'll keep looking at it when I'm
back tomorrow.

Neverlord:
@timwoj thanks for bringing back MSVC builds! I have updated the ALM branch and #201 has (working) Windows builds back as well. Closing here.

PR #214, opened 2022-03-02
Rework CI scripts for Windows

timwoj:
This PR reworks all of the CI scripts for Windows builds to properly produce a build. The biggest change is the switch using the versions of CMake and Ninja that can be installed via MSVC.

Note that the build currently fails, but that issue is being tracked via #213.

### DIFF
--- a/ci/windows/build.cmd
+++ b/ci/windows/build.cmd
@@ -1,9 +1,14 @@
+:: Import the visual studio compiler environment into the one running in the
+:: cmd current shell. This path is hard coded to the one on the CI image, but
+:: can be adjusted if running builds locally. Unfortunately, the initial path
+:: isn't in the environment so we have to hardcode the whole path.
 call "c:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64
 
 mkdir build
 cd build
 
-:: All of the relevant binaries here are put in the path by the bat file above
+:: The configure script doesn't work for Windows but we can call cmake directly
+:: to do the same. Ninja is installed as part of the CI image.
 C:\WINDOWS\system32\cmd.exe /c %SYSTEMROOT%\System32\chcp.com 65001 >NUL && ^
 cmake.exe -G "Ninja" -DCMAKE_BUILD_TYPE:STRING="Debug" ^
 -DCMAKE_C_COMPILER=cl.exe -DCMAKE_CXX_COMPILER=cl.exe -DCMAKE_MAKE_PROGRAM=ninja.exe ^
--- a/ci/windows/build.cmd
+++ b/ci/windows/build.cmd
@@ -1,6 +1,13 @@
-set PATH=%PATH%;C:\Program Files\CMake\bin
+call "c:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64
+
 mkdir build
 cd build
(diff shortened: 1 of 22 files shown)


Neverlord:
@timwoj I took the liberty to push to your branch. :)

Should be good to go, except that `ctest` is currently missing on the Windows node.

timwoj:
> @timwoj I took the liberty to push to your branch. :)

No, that's awesome. I'd much rather have it here all as one thing! I'll take a look at getting ctest onto the VM this week.

timwoj:
I made a few more changes. The python tests are all failing because of this:

```
Traceback (most recent call last):
  File "C:\broker\build\python\broker\__init__.py", line 3, in <module>
    from . import _broker
ImportError: cannot import name '_broker' from partially initialized module 'broker' (most likely due to a circular import) (C:\broker\bui
ld\python\broker\__init__.py)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\broker\tests\python\communication.py", line 8, in <module>
    import broker
  File "C:\broker\build\python\broker\__init__.py", line 5, in <module>
    import _broker
ModuleNotFoundError: No module named '_broker'
```

We really ought to consider enabling verbose output by default, at least for CI runs. It'd be more useful. I'm working on the `generator_file_writer` and `meta_command_writer` test failures. There's an issue when building related to a macro it generates being wrong for MSVC, so I'm hoping it's that.

timwoj:
I went ahead and disabled the python bindings in Windows builds temporarily in the interest of landing all of the rest of these fixes. I'll open a separate issue on fixing those.

ci/windows/build.cmd
```diff
@@ -1,6 +1,14 @@
-set PATH=%PATH%;C:\Program Files\CMake\bin
+call "c:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64
```
ckreibich, line 1: Just a minor suggestion: comment these ultra-windowsy bits (also below) a bit more?  You're saying this puts stuff in the path, but maybe there's a handy docs link for it or some such, for when we need to revisit. Etc.

Any chance there's a way to make the path to it less hardwired? (Perhaps that's normal, I really don't know this stuff!)

include/broker/worker.hh
```diff
@@ -16,6 +17,16 @@ public:
 
   struct impl;
 
+  // -- constants --------------------------------------------------------------
+
+#ifdef BROKER_WINDOWS
```
ckreibich, line 22: Explain the `*2` with a comment?

ckreibich (approved):
I've only done pattern-matching and jaw-dropping on this because I'm not set up to build/test this, but at a high level this all looks good to me. Kudos for figuring all of this out. My main suggestion is to comment the Windows-heavy bits a bit more to give some context. Maybe it's clear as-is for Windows folks, but it seems we have a high probability of non-Windows people needing to touch this for the foreseeable future.

include/broker/worker.hh
timwoj: I don't actually know the reasoning for this one. @Neverlord do remember why this is?

Neverlord: Ups, sorry for the delayed response. 😅

The `*2` is just a cheap workaround for now. We use placement new to construct a `caf::actor` handle into the buffer. This type is just a wrapper around a pointer and both GCC and Clang agree that `sizeof(caf::actor) == sizeof(void*)`. However, MSVC seems to add some padding. There might be ways to tell MSVC to produce a compact struct, but since `caf::actor` is not in the Broker code base I've opted for this workaround.

I didn't bother documenting it since the whole placement-new approach is ugly and meant as a temporary solution. It was the best thing to do without changing CAF. However, the much better solution is to add a getter to `caf::actor` to retrieve the underlying pointer. I plan on adding such a getter to CAF for the next release and then update the handle implementations in Broker.

PR #219, opened 2022-04-30
Add new WebSocket interface to Broker

Neverlord:
Allow clients to connect to Broker without having to implement the participate in the peer-to-peer communication. Clients connect via WebSocket and exchange JSON-formatted text messages with the Broker endpoint.

- [x] Add configuration option to open a WebSocket server.
- [x] Implement initial version of the JSON client protocol.
- [x] Document the JSON protocol.
- [x] ~~Provide a JSON schema for the protocol.~~ (let's add that later, after the merge)
- [x] Implement test suite that exercises the WebSocket protocol with several edge cases (like error reporting).
- [x] Support encrypted WebSocket communication (wss).


Neverlord:
@rsmmr I haven't relaxed the timestamp parser yet, but I've put some code in to prevent the other two issues you've reported. Let me know if these issues still pop up in your testing. The server also sends error messages to clients now, so at least you get some form of feedback if the server drops invalid input. There are three types of messages that the server can send now:
- ack: this is the first message that the server sends *after* it receives the filter; contains the Broker version and the endpoint ID. This message appears only once.
- event: this is a `data_message` represented as before with topic and data.
- error: contains a human-readable code (name of `broker::errc`) with some additional context if available.

I have also implemented a first high-level [Robot](https://robotframework.org) test under `tests/robot` that spins up a Broker endpoint (using the `broker-node` binary) with WebSocket enabled and then attaches two clients: one that publishes and one that receives. Let me know what you think of that. If you're not keen on using Robot we could use something else instead. The first test is pretty basic. If we stick to Robot for this, I would add additional tests that check for correct filtering and error reporting.

rsmmr:
Thanks, will try it out.

> I have also implemented a first high-level [Robot](https://robotframework.org) test under `tests/robot` that spins up a Broker endpoint (using the `broker-node` binary) with WebSocket enabled and then attaches two clients: one that publishes and one that receives. Let me know what you think of that. If you're not keen on using Robot we could use something else instead. The first test is pretty basic. If we stick to Robot for this, I would add additional tests that check for correct filtering and error reporting.

I'd prefer to not add another testing framework as that would need to be installed/maintained across CI etc. It's a trade-off of course in terms of how much it would gain us to use Robot vs coding tests differently. How difficult would it be to do the tests, say, in plain Python and execute through btest (or whatever other way you'd prefer using existing machinery)?

Neverlord:
Robot is pure python and I have had good experience with it recently. It's just a `pip install` away, but it seems the same goes for `btest`. Using `btest` just didn't occur to me TBH, it's "the Zeek testing infrastructure" in my mind. 🙂

As long as we stick to baseline checks, we can probably do the tests with `btest` with comparable effort. Let me re-implement that test with `btest` and see if there's any major downside compared to Robot.

rsmmr:
> As long as we stick to baseline checks, we can probably do the tests with `btest` with comparable effort. Let me re-implement that test with `btest` and see if there's any major downside compared to Robot.

Sounds good. if it's similar work, I prefer `btest`  because "everything Zeek" will already have it installed, and there's no learning curve for anybody. (Just as a note, btest doesn't necessarily mean/require "baseline"; one can also signal failure simply through a script exiting with failure; whatever works best).

rsmmr:
> Thanks, will try it out.

Confirming that the changes all work well for me (the two fixes, and the new message types).

Neverlord:
@rsmmr I have added a new manual section on the WebSocket JSON API (feedback welcome) and I have also streamlined the JSON format: the top-level "data" for event messages has been stripped. This probably breaks your existing code, but I think it makes much more sense this way. See the new manual section for an example event. 😉

I have also changed/fixed the encoding of `subnet`. This was broken before for IPv4 subnets, because Broker would internally convert the address to v6 and then use the length as-is. Meaning that the subnet for any v4 address was off by 96. JSON uses the "slash-notation" now, which is IMO superior anyway.

Neverlord:
This is done, only the btest suites are failing on some builds. I'll clean that up, but aside from getting CI green, this is ready for review.

doc/web-socket.rst
```diff
@@ -0,0 +1,405 @@
...
+==========
+
+Broker offers access to the publish/subscribe layer via WebSocket in order to
+make its :ref:`data <data-model>` accessible to third parties.
+
+WebSocket clients are treated as lightweight peers. Each Broker endpoint can be
+configured to act as a WebSocket server by either setting the environment
+variable ``BROKER_WEB_SOCKET_PORT`` or by setting ``broker.web-socket.port`` on
+the command line or in the configuration file. When running Zeek, scripts may
+also redefine ``Broker::web_socket_port`` to a valid TCP port.
```
rsmmr, line 13:
```suggestion
also redefine ``Broker::default_port_websocket`` to a valid TCP port.
```

That's how my Zeek-side branch now names this.

doc/web-socket.rst
```diff
@@ -0,0 +1,405 @@
...
+
+Protocol
+~~~~~~~~
+
+After the handshake, the WebSocket client may only send `Data Messages`_. The
+Broker endpoint converts every message to its native representation and
+publishes it.
+
+The WebSocket server may send `Event Messages`_ (whenever a data message matches
+the subscriptions of the client) and `Error Messages_` to the client.
```
rsmmr, line 65: I think the distinction between data messages and event message isn't clear here. If we are in pure Broker-land (no Zeek), aren't then both sides only sending generic data messages? I.e., no event semantics. If we are talking Zeek, then both sides are sending events, which have a specific format. I would suggest to make that distinction explicit: talk about the two cases and (1) show generic messages in the first case, and (2) show concrete examples of how a Zeek event looks liken the 2nd. The latter will then also help people writing Zeek clients to know the event format, because that's currently not documented anywhere I believe (on the C++ side, we got a wrapper hiding that iirc).

Additional thought: maybe using `event` as the message type isn't even quite right at the Broker layer? And more generally, can we make the two sides symmetrical so that client and server send their data in the same format (data messages I guess)?

doc/web-socket.rst
```diff
@@ -0,0 +1,405 @@
...
+Boolean
+*******
+
+The type ``boolean`` can take on exactly two values and maps to the native JSON
+boolean type:
+
+.. code-block:: json
+
+  {
+    "@data-type": "boolean",
```
rsmmr, line 101: Why's this prefixed with `@`, whereas other keys aren't? (Just curious).

rsmmr (commented):
Looked over the code, and to the degree I can tell, I didn't spot anything particular. Left some notes on the docs.

This also needs rebase to `master`, doesn't apply cleanly.

rsmmr:
There's a parallel PR for Zeek: https://github.com/zeek/zeek/pull/2121 (still WIP)

doc/web-socket.rst
Neverlord: The idea is to flag meta fields with an `@` since they are more like annotations than actual fields. In this case, this is an annotation for the `data` field that tells the parser what type to use.

doc/web-socket.rst
Neverlord: I've changed this in [5b26424](https://github.com/zeek/broker/pull/219/commits/5b26424a823fb12cda46e9e009bb97e6ce97c793).

doc/web-socket.rst
rsmmr: Looks good, I adapted the agent and added a Zeek-side test in https://github.com/zeek/zeek/pull/2121. All working well.

Neverlord:
@rsmmr I believe this should be good to go. I've added documentation for how Broker renders Zeek events, renamed the path to `/v1/messages/json` and made the message types symmetrical. Let me know if I forgot something. 🙂

doc/web-socket.rst
```diff
@@ -404,3 +404,66 @@ message would cause it to send this error back to the client:
...
+
+``type``
+  A ``count`` denoting the encoded Zeek message type. For events, this is always
+  ``1``. Other message types in Zeek are currently not safe for 3rd-party use.
+
+``name``
+  Identifies the Zeek event.
+
+``args``
+  Contains the arguments for the event. Usually, this is another ``vector``.
```
rsmmr, line 425:
```suggestion
  Contains the arguments for the event in the form of another ``vector``.
```

rsmmr (approved):
Looks good. 

Let me know if anybody else wants to take a look, otherwise I'll go ahead and merge (probably on Monday).

PR #230, opened 2022-05-29
Set SO_REUSEADDR by default

Neverlord:
Note: This PR forks off from #219 to add the new `reuse_addr` flag to the native as well as to the WebSocket `listen` function in one go. The first actual commit for this PR is e95cfc522ebccfcf9b61824a30f61fae4117b9c7.

@jlucovsky @rsmmr The original patch had users explicitly enable this flag and a custom type to do so (and also had the option to set `SO_REUSEPORT`). This is a simplified version with just a boolean flag. After reviewing the original patch and googling a bit more, I think there's not really a downside to enabling reuse-addr by default and not caring about `SO_REUSEPORT`. Let me know if you have different opinions. 🙂


Neverlord:
Oh, forgot to mention: this API change is backwards-compatible since all new parameters have default values.

Issue #246, opened 2022-06-24
Some tests seem unstable on Fedora


Neverlord:
I had it a couple times now that a CI run on Fedora would turn red. Restarting it usually turns it back to green, but we should make the tests stable of course. Not sure yet if these errors point to an actual issue in Broker or 'just' affect the test code. Working on it.

Neverlord:
Huh, one recent run had this error:

```
/root/.local/bin/btest-diff: line 224: diff: command not found
```

I don't understand why re-running solved this, because either our Docker image installs the necessary package or it doesn't. 🤔

Neverlord:
Seems like this is fixed now, at least I couldn't see a CI error since merging the PR. I'll leave this ticket open for a bit and close it if CI consistently passes from now on.

Neverlord:
Didn't see this error show up on CI again after the merge. Seems fixed.

PR #247, opened 2022-06-25
Fix double-close on pipe handles

Neverlord:
I have tracked down some infrequent unit tests on Fedora (see https://github.com/zeek/broker/issues/246) to double-close on the pipe handles in the `connector`.

The `connector` closes its pipe handles last, but the `connect_manager` closes them beforehand by blindly closing all sockets in its pollset. The error happens only in our multi-threaded tests:

- Thread A closes the pipe handle X (`~connect_manager`)
- Thread B accepts a new connection and the OS chooses X since it's just been released
- Thread A runs `~connector`, which closes X *again*
- Thread B gets an EBADFD seemingly out of nowhere and rightfully complains

I think the connector manager should not close any sockets at all, since these sockets are owned by other components that take ownership.

With this patch, I could run the `system.peering` suite on Fedora several hundred times without error during testing. Before the patch, it would break every couple dozen runs on average. Also, this has nothing to with Fedora and is simply a race on the Kernel's socket table. Just happened to occur most frequently in our Fedora build.

### DIFF
--- a/src/internal/connector.cc
+++ b/src/internal/connector.cc
@@ -983,17 +983,6 @@ struct connect_manager {
 
   connect_manager& operator=(const connect_manager&) = delete;
 
-  ~connect_manager() {
-    for (auto& entry : fdset) {
-      BROKER_DEBUG("close socket" << entry.fd);
-      caf::net::close(caf::net::socket{entry.fd});
-    }
-    for (auto& entry : pending_fdset) {
-      BROKER_DEBUG("close socket" << entry.fd);
-      caf::net::close(caf::net::socket{entry.fd});
-    }
-  }
-
   /// Returns the relative timeout for the next retry in milliseconds or -1.
   int next_timeout_ms() {
     if (retry_schedule.empty())


timwoj:
Is this a big enough problem that I should bring this over into 5.0?

Neverlord:
> Is this a big enough problem that I should bring this over into 5.0?

Good question. It's a shutdown issue and the bug mostly manifests when (re-) starting multiple Broker endpoints in the same process. For Zeek itself, this is probably unlikely to cause issues.

Neverlord:
@timwoj CI failed again because of `/root/.local/bin/btest-diff: line 224: diff: command not found` (just restarted it). So far, I think I've seen this only on Fedora 34. Since https://github.com/zeek/broker/pull/239 drops that image, I'd wait after both PRs landed to see whether there still needs something to be done.

timwoj:
That's a weird one. I wonder if adding diffutils to the Fedora 34 Dockerfile would permanently fix it. I'm not sure why it would work sometimes but not other times.

Neverlord:
@timwoj Ok, then I'll rebase this PR after you've merged in #239. Should jump back to green afterwards.

timwoj:
> Ok, then I'll rebase this PR after you've merged in https://github.com/zeek/broker/pull/239. Should jump back to green afterwards.

239 is merged now.

Neverlord:
@timwoj rebased and ready to go. 🙂

Neverlord:
> I merged this already but for some reason GitHub didn't close the PR.

Looks like you've merged an older commit ([1582461](https://github.com/zeek/broker/commit/15824610a67e345656890233eb26c4eeb45f1d35) instead of [0088497](https://github.com/zeek/broker/commit/0088497ac77e7db17e60a56345ad26c12e33081c)). That's fine, the rebase didn't change any code, but GitHub does not recognize it because the PR points to a never commit.

PR #248, opened 2022-06-28
Make more and better use of C++17 features

Neverlord:
This is PR bundles a bunch of refactoring that I've meant to do for a bit now and finally got around doing. I'm not particularly happy with the `convert` API. It 'injects' free functions like `to` and `to_string`, but it's not really ADL-safe and only works if the target type is in the namespace `broker`. Using an interface like we do for [`comparable`](https://github.com/zeek/broker/blob/master/include/broker/detail/comparable.hh) would be much more robust, so I might revisit that at some point.

For now, this PR focuses on making better use of the tools we have access to since switching to C++17:

- Use `if constexpr` over `enable_if` in more places
- Consequently use `foo_v<...>` over `foo<...>::value`
- Drop legacy C++14 and C++17 backports in `type_traits.hh`
- Use better type introspection to allow `convert` to return `void`
- Make more consistent use of the `convert` API


rsmmr (commented):
> Use better type introspection to allow convert to return void

Is this change applying only to the "to string" direction, where we know that conversion now can never fail anymore?

From an API perspective, this is something that user code might be using, right? So we might end up breaking code here?

General note for the future: it would be quite helpful to have semantically independent changes come as separate commits, that would make reviewing a PR like this easier.

rsmmr:
> The convert functions aren't meant for user code. 

Ack, ok. Fine with me to merge this then.

rsmmr (approved):

Issue #249, opened 2022-06-29
CI: Run clang-tidy, clang-format, cppcheck


Neverlord:
This was originally a task in #212. Basically, we want to add additional quality gates to our CI:

- [x] `clang-format` to make sure we use a consistent coding style
- [x] `clang-tidy` for making sure the code avoids common pitfalls (`bugprone` checks) and follows some best practices for readability
- [ ] `cppcheck` to have an additional static analyzer look over the source code (Zeek's Coverity runs also cover Broker)

PR #255, opened 2022-07-09
Apply clang-format to the code base

Neverlord:
- Apply clang-format to the entire code base
- Enable clang-format checks for PRs

Aside from the obvious benefit of forcing a consistent coding style, this PR is also a preparation step for running `clang-tidy`. Some `clang-tidy` checks can automatically provide a fix for certain findings. However, when applying such tool-assisted code changes, we also need to run `clang-format` afterwards to make sure the code stays consistently formatted.


Neverlord:
I've ended up excluding the code under `tests/benchmark/readerwriterqueue/`. The diff had some very weird looking code changes and that code is very unlikely to change. Maybe removing that code is actually the better call, but I'd prefer doing that separately.

bindings/python/set_bind.h
```diff
@@ -9,142 +10,145 @@
...
 
-/* For a set data structure, recursively check the value type (which is std::pair for maps) */
+/* For a set data structure, recursively check the value type (which is
+ * std::pair for maps) */
 template <typename T>
 struct is_comparable<T, enable_if_t<container_traits<T>::is_set>> {
-    static constexpr const bool value =
-        is_comparable<typename T::value_type>::value;
+  static constexpr const bool value
+    = is_comparable<typename T::value_type>::value;
```
timwoj, line 18: I think I prefer the `=` on the line above.

bindings/python/set_bind.h
```diff
@@ -9,142 +10,145 @@
...
-template <typename, typename, typename... Args> void set_if_insertion_operator(const Args &...) { }
-template <typename, typename, typename... Args> void set_modifiers(const Args &...) { }
-
-template<typename Set, typename Class_>
-void set_if_copy_constructible(enable_if_t<
-    std::is_copy_constructible<Set>::value &&
-    std::is_copy_constructible<typename Set::value_type>::value, Class_> &cl) {
+template <typename, typename, typename... Args>
+void set_if_copy_constructible(const Args&...) {
+}
```
timwoj, line 24: Prefer to have empty brace sets on the same line.

bindings/python/_broker.cpp
```diff
@@ -169,134 +173,168 @@ PYBIND11_MODULE(_broker, m) {
     .def("capacity", &broker::publisher::capacity)
     .def("fd", &broker::publisher::fd)
     .def("drop_all_on_destruction", &broker::publisher::drop_all_on_destruction)
-    .def("publish", (void (broker::publisher::*)(broker::data d)) &broker::publisher::publish)
-    .def("publish_batch",
-       [](broker::publisher& p, std::vector<broker::data> xs) { p.publish(xs); })
+    .def("publish", (void(broker::publisher::*)(broker::data d))
+                      & broker::publisher::publish)
```
timwoj, line 177: I'm not sure why this is separating the `&` from the variable name, if the intention is for it to be taking the address of that variable.

doc/_examples/pong.cc
```diff
@@ -9,28 +9,28 @@
...
-    auto ss = ep.make_status_subscriber(true);
-    ep.listen("", 9999);
+  // Setup endpoint and connect to Zeek.
+  endpoint ep;
+  auto sub = ep.make_subscriber({"/topic/test"});
+  auto ss = ep.make_status_subscriber(true);
+  ep.listen("", 9999);
 
-    // Wait until connection is established.
-    for ( bool has_peer = false; !has_peer; ) {
```
timwoj, line 19: This was a weird mix of the Zeek style and whatever broker uses by default. There's actually a flag you can pass to clang-format to preserve this style, if that's what we want to do (`SpacesInConditionalStatement`). It doesn't look like we prefer this format everywhere else though, so I'm fine with letting it go.

include/broker/detail/filesystem.hh
```diff
@@ -8,7 +8,7 @@
 
 #ifdef BROKER_HAS_STD_FILESYSTEM
 
-#include <filesystem>
+#  include <filesystem>
```
timwoj, line 11: Is this because of the macro indentation rules?

include/broker/detail/radix_tree.hh
```diff
@@ -699,8 +699,8 @@ radix_tree<T, N>::find_child(node* n, unsigned char c) {
 }
 
 template <typename T, std::size_t N>
-size_t radix_tree<T, N>::prefix_shared(node* n, const key_type& key,
-                                       int depth) {
+size_t
+radix_tree<T, N>::prefix_shared(node* n, const key_type& key, int depth) {
```
timwoj, line 703: I'm not sure how I feel about this one. I think I would prefer the old way here, breaking the line in the argument list instead of after the return type.

tests/cpp/core.cc
```diff
@@ -28,8 +28,16 @@ struct driver_state {
   buf_type xs;
   static inline const char* name = "driver";
   void reset() {
-    xs = data_msgs({{"a", 0}, {"b", true}, {"a", 1}, {"a", 2}, {"b", false},
-                    {"b", true}, {"a", 3}, {"b", false}, {"a", 4}, {"a", 5}});
+    xs = data_msgs({{"a", 0},
```
timwoj, line 31: Weird. i wonder why in every other case of a construct like this, it compacted the list into fewer lines, but in this case it expanded it.

timwoj (changes requested):
I made it through the whole thing. One general comment is that we're missing a `.clang-format` file. Even if we're just using one of the default styles, we probably should have one just to be explicit about what style we're opting for.

src/internal/master_actor.cc
```diff
@@ -460,7 +460,7 @@ caf::behavior master_state::make_behavior() {
   BROKER_TRACE(BROKER_ARG(id) << BROKER_ARG(core) << BROKER_ARG(store_name));
   // Setup.
   self->monitor(core);
-  self->set_down_handler([this](const caf::down_msg& msg) {
+  self->set_down_handler([this](const caf::down_msg& msg) { //
```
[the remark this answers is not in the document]
Neverlord: I've added this manually to force clang-format to break it here. Otherwise, clang-format breaks before the lambda and adds a weird extra indentation level that makes it harder to read. There are actually a couple places I do this in the code base.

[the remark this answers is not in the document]
timwoj: Ah, right on.

bindings/python/_broker.cpp
Neverlord: I'm not sure either was `clang-format` is thinking about this code. Maybe it confuses this with an argument declaration. Although, then it should put the `&` after the brace... 🤷🏻‍♂️

include/broker/internal/type_id.hh
```diff
@@ -166,12 +166,12 @@ CAF_BEGIN_TYPE_ID_BLOCK(broker, caf::first_custom_type_id)
 
   // -- STD/CAF type announcements ---------------------------------------------
 
-  BROKER_ADD_TYPE_ID((std::optional<broker::endpoint_id>))
-  BROKER_ADD_TYPE_ID((std::optional<broker::timespan>))
-  BROKER_ADD_TYPE_ID((std::optional<broker::timestamp>))
-  BROKER_ADD_TYPE_ID((std::shared_ptr<broker::filter_type>))
-  BROKER_ADD_TYPE_ID((std::shared_ptr<std::promise<void>>))
-  BROKER_ADD_TYPE_ID((std::vector<broker::peer_info>))
+  BROKER_ADD_TYPE_ID((std::optional<broker::endpoint_id>) )
```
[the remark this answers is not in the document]
Neverlord: I have no idea either. The space between the braces makes zero sense to me (and shouldn't be there) but I didn't investigate why `clang-format` gets confused here.

tests/cpp/backend.cc
```diff
@@ -203,13 +169,13 @@ struct fixture : base_fixture {
...
 
-} // namespace <anonymous>
+} // namespace
 
 FIXTURE_SCOPE(backend_tests, fixture)
 
 #define RUN(statement) run([&] { return statement; }, #statement)
 
-TEST(put/get) {
+TEST(put / get) {
```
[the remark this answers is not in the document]
Neverlord: Well, `clang-format` is right to put whitespaces around binary operators. And from this syntax it definitely has to assume that this is a division (or using some overloaded operator). The problem here is that the macro silently converts the name into a string when it really should just accept a string instead. That actually annoys me for a long time now, but not enough (yet) to go through the motions of updating the macro and updating all call sites.

tests/cpp/core.cc
Neverlord: I wondered about this particular formatting change as well, because `clang-format` usually seems to try and minimize line count. Though it's well readable that way, IMHO.

Neverlord:
@timwoj I've tweaked the coding style based on your suggestions:
- keep `=` on the line above
- allow empty functions on a single line

I've also tweaked the penalties so that `clang-format` generally tries to avoid putting the return type on a single line and I had some cases where `clang-format` would break in the middle of a structured binding after putting `=` on the line above without some tuning (setting `PenaltyBreakAssignment` to 21 is the sweetspot apparently).

Let me know if it needs further tweaking. If you prefer, I can also squash the commits before a merge (kept the second commit for now to allow you to review the changes).

Neverlord:
Also, should've mentioned that his PR relates #249.

timwoj (approved):
Looks good to me now. clang-format is just weird sometimes, so the rest of the questions I had can just stay the way they are.

PR #258, opened 2022-07-28
Add support metric_registry_impl::collect

awelzel:
...also add or make some of the free functions take `const` values.

These changes are being used in the following WIP PR for Zeek: https://github.com/zeek/zeek/pull/2273

### DIFF
--- a/include/broker/telemetry/fwd.hh
+++ b/include/broker/telemetry/fwd.hh
@@ -32,7 +32,7 @@ class histogram_family;
 
 using label_view = std::pair<std::string_view, std::string_view>;
 
-using label_list = span<label_view>;
+using const_label_list = span<const label_view>;
 
 using dbl_counter = counter<double>;
 
--- a/include/broker/telemetry/fwd.hh
+++ b/include/broker/telemetry/fwd.hh
@@ -90,7 +90,7 @@ void inc(dbl_counter_hdl*) noexcept;
 
 void inc(dbl_counter_hdl*, double amount) noexcept;
 
-double value(dbl_counter_hdl*) noexcept;
+double value(const dbl_counter_hdl*) noexcept;
 
 dbl_counter_hdl* dbl_counter_get_or_add(metric_family_hdl*,
                                         span<const label_view> labels);
@@ -99,7 +99,7 @@ int64_t inc(int_counter_hdl*) noexcept;
 
 void inc(int_counter_hdl*, int64_t amount) noexcept;
 
-int64_t value(int_counter_hdl*) noexcept;
+int64_t value(const int_counter_hdl*) noexcept;
 
 int_counter_hdl* int_counter_get_or_add(metric_family_hdl*,
                                         span<const label_view> labels);
@@ -114,7 +114,7 @@ void dec(dbl_gauge_hdl* hdl, double amount) noexcept;
 
 void dec(dbl_gauge_hdl* hdl) noexcept;
 
-double value(dbl_gauge_hdl* hdl) noexcept;
+double value(const dbl_gauge_hdl* hdl) noexcept;
 
 dbl_gauge_hdl* dbl_gauge_get_or_add(metric_family_hdl* hdl,
                                     span<const label_view> labels);
@@ -127,7 +127,7 @@ int64_t dec(int_gauge_hdl* hdl) noexcept;
 
 void dec(int_gauge_hdl* hdl, int64_t amount) noexcept;
 
-int64_t value(int_gauge_hdl* hdl) noexcept;
+int64_t value(const int_gauge_hdl* hdl) noexcept;
 
 int_gauge_hdl* int_gauge_get_or_add(metric_family_hdl* hdl,
                                     span<const label_view> labels);
@@ -136,26 +136,26 @@ int_gauge_hdl* int_gauge_get_or_add(metric_family_hdl* hdl,
 
 void observe(dbl_histogram_hdl*, double value) noexcept;
 
-double sum(dbl_histogram_hdl*) noexcept;
+double sum(const dbl_histogram_hdl*) noexcept;
 
-size_t num_buckets(dbl_histogram_hdl*) noexcept;
+size_t num_buckets(const dbl_histogram_hdl*) noexcept;
 
-double count_at(dbl_histogram_hdl*, size_t index) noexcept;
+double count_at(const dbl_histogram_hdl*, size_t index) noexcept;
 
-double upper_bound_at(dbl_histogram_hdl*, size_t index) noexcept;
+double upper_bound_at(const dbl_histogram_hdl*, size_t index) noexcept;
 
 dbl_histogram_hdl* dbl_histogram_get_or_add(metric_family_hdl*,
                                             span<const label_view> labels);
 
 void observe(int_histogram_hdl*, int64_t value) noexcept;
 
-int64_t sum(int_histogram_hdl*) noexcept;
+int64_t sum(const int_histogram_hdl*) noexcept;
 
-size_t num_buckets(int_histogram_hdl*) noexcept;
+size_t num_buckets(const int_histogram_hdl*) noexcept;
 
-int64_t count_at(int_histogram_hdl*, size_t index) noexcept;
+int64_t count_at(const int_histogram_hdl*, size_t index) noexcept;
 
-int64_t upper_bound_at(int_histogram_hdl*, size_t index) noexcept;
+int64_t upper_bound_at(const int_histogram_hdl*, size_t index) noexcept;
 
 int_histogram_hdl* int_histogram_get_or_add(metric_family_hdl*,
                                             span<const label_view> labels);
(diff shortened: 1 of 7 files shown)


src/telemetry/metric_registry.cc
```diff
@@ -185,6 +185,82 @@ class impl_base : public metric_registry_impl {
...
+      return r;
+    }
+
+    metrics_collector& collector_;
+  };
+
+  void collect(metrics_collector& collector) {
+    auto proxy_collector = ProxyCollector(collector);
+    reg_->collect(proxy_collector);
+  }
```
Neverlord, line 262: The `operator()` overloads invoke UB, e.g., by using a `reinterpret_cast` from `ct::metric_family` to `dbl_counter_family_hdl` (which is the type-erased version of `ct::metric_family_impl<dbl_counter>`). This "works" on most current platforms, but only if there arent' any virtual base types involved.

In any case, the proxy class is indeed unnecessary and adds a lot of boilerplate code. Could just use a couple of free function helpers and a lambda:

```cpp
std::vector<std::string_view> extract_label_values(const ct::metric* instance) {
  auto get_value = [](const auto& lbl) {
    auto val = lbl.value();
    return std::string_view{val.data(), val.size()};
  };
  auto labels = instance->labels();
  std::vector<std::string_view> result;
  std::transform(labels.begin(), labels.end(), std::back_inserter(result), get_value);
  return result;
}
    
const auto* opaque(const ct::metric_family_impl<ct::dbl_counter>* family) {
  return reinterpret_cast<const dbl_counter_hdl*>(family);
}

// .. some more opaque overloads, could also be macro-generated ..
```

With these, the `collect` function becomes straightforward:

```cpp
  void collect(metrics_collector& collector) {
    auto fn = [&collector](const auto* family, const ct::metric* instance, const auto* obj) {
      auto labels = extract_label_values(instance);
      collector(opaque(family), opaque(obj), labels);
    };
    reg_->collect(fn);
  }
```

(Please also note my comment regarding extracting the label values only. So probably the free function would be `extract_labels` and assemble the key-value pairs. It's probably also a good idea to pass the vector the `extract` function and re-fill the same vector over and over again instead of allocating fresh memory each time.)

include/broker/telemetry/metric_registry_impl.hh
```diff
@@ -9,6 +9,37 @@
...
 namespace broker::telemetry {
 
+// Collector for the broker::telemetry layer. We string through the label
+// values of individual metric instances as there aren't direct accessors
+// for those from a hdl.
+class metrics_collector {
+public:
+  virtual void operator()(const dbl_counter_family_hdl* family,
+                          const dbl_counter_hdl* counter,
+                          const std::vector<std::string_view>& label_values)
```
Neverlord, line 19: 1. Having only the values here makes it very cumbersome to correlate them back to the keys. Basically forces users to go back to the `family` and rely on index lookups.
2. Forcing the consumer to care about `std::vector` here also seems unnecessary.

If you prefer not lifting CAF's `label` type, this should at least least become `label_list labels`, with:

```cpp
  /// A list of key-value pairs representing the label dimensions of a metric.
  using label_list = span<const std::pair<std::string_view, std::string_view>>;
```

Neverlord (changes requested):
Overall, this looks like the right approach. I think the interface of `metrics_collector` is a bit odd (only giving access to the label values) and the implementation can be streamlined (plus it currently has UB by using `reinterpret_cast` to perform a downcast).

src/telemetry/metric_registry.cc
awelzel: I've followed your suggested route with using lambda and free-functions, but just for clarification:

> The operator() overloads invoke UB, e.g., by using a reinterpret_cast from ct::metric_family to dbl_counter_family_hdl (which is the type-erased version of ct::metric_family_impl<dbl_counter>). This "works" on most current platforms, but only if there arent' any virtual base types involved.

This would have been "better" if the individual `operator()` implementations had used concrete `ct::dbl_counter_family *` rather than `ct::metric_family *` (as the `opaque()` helpers now will take?).

The docs and the prom collector which I had copied use the `ct::metric_family *` - that's where this was coming from:
https://github.com/actor-framework/actor-framework/blob/master/manual/Metrics.rst#the-metric-registry

awelzel: I've pushed the changes. Thanks, it got shorter and looks better.

TBH, I'm slightly intimidated by that lambda auto type deduction ;-)

For the final merge I'll squash the commits together.

Mind giving it another look? Thanks!

Neverlord: > This would have been "better" if the individual `operator()` implementations had used concrete `ct::dbl_counter_family *` rather than `ct::metric_family *` (as the `opaque()` helpers now will take?).

Yes. By taking the argument as `ct::metric_family*`, you're forcing a downcast to the base type. For going back to the concrete type, you need to use `static_cast` (if you're 100% certain you're casting to the right type) or `dynamic_cast`. Using a `reinterpret_cast` to do this is unsafe.

> The docs and the prom collector which I had copied use the ct::metric_family * - that's where this was coming from:

The CAF example does not care about the concrete type of the family objects. You generally only need those for having access to `get_or_add` (i.e., navigate to or create new instances), but while you are iterating the registry, you'll have `const` access anyway. When collecting metrics, the base type is usually all you need for the family. The new `metrics_collector` could've done the same, but then the `operator()` overloads should take `metric_family_hdl*` as first argument in all overloads. The question really is whether you see a use case for having the concrete family available or not. If not, you could drop half the `opaque` overloads (for the family types) and streamline the code a bit. :)

include/broker/telemetry/fwd.hh
```diff
@@ -32,6 +32,8 @@ class histogram_family;
 
 using label_view = std::pair<std::string_view, std::string_view>;
 
+using label_list = span<label_view>;
```
Neverlord, line 35: I think restricting access to `const` is better suited for the use cases.

```suggestion
using const_label_list = span<const label_view>;
```

src/telemetry/metric_registry.cc
```diff
@@ -107,6 +109,50 @@ metric_registry::~metric_registry() {
...
+    auto val = label.value();
+    return {std::string_view{name.data(), name.size()},
+            std::string_view{val.data(), val.size()}};
+  };
+  vec.clear();
+  const auto& labels = instance->labels();
+  std::transform(labels.begin(), labels.end(), std::back_inserter(vec),
+                 get_value);
+
+  return vec;
```
Neverlord, line 127: Having an output parameter *and* returning it feels quite redundant (and can obfuscate the call side). I think this should simply return `void`.

```suggestion
```

src/telemetry/metric_registry.cc
```diff
@@ -185,6 +231,19 @@ class impl_base : public metric_registry_impl {
...
   }
 
+  void collect(metrics_collector& collector) override {
+    std::vector<label_view> labels_vec; // Reuse for label extraction
+    auto fn = [&collector, &labels_vec](const auto* family,
+                                        const ct::metric* instance,
+                                        const auto* obj) {
+      labels_vec.clear();
+      auto labels = extract_labels(instance, labels_vec);
+      collector(opaque(family), opaque(obj), labels);
```
Neverlord, line 241:
```suggestion
      extract_labels(instance, labels_vec);
      collector(opaque(family), opaque(obj), labels_vec);
```

Neverlord (changes requested):
Two nitpicks, but other than that looks good now!

src/telemetry/metric_registry.cc
awelzel: I've did as you suggested. The Zeek consumer only did upcast. I did keep the const `upcast()` changes in this PR. If you'd like to see them go, can remove them.

I've addressed the other two nitpicks as well.

Thanks again.

awelzel: Ah - I'm not sure what more streamlining there could've been done.

Neverlord (approved):
LGTM!

One last (very minor) comment: since we no longer care about the actual `family` type, the lambda could take the `family` argument as `const ct::metric_family* family` to avoid unnecessary template parameters.

> For the final merge I'll squash the commits together.

Unless @timwoj has a different opinion on this, I personally wouldn't squash since it messes with the discussions we had here.

src/telemetry/metric_registry.cc
Neverlord: Sorry, that was unclear. I just meant getting rid of the macro stuff for the family overloads to streamline the code. Looks good as-is now.

awelzel:
> One last (very minor) comment: since we no longer care about the actual family type, the lambda could take the family argument as const ct::metric_family* family to avoid unnecessary template parameters.

Got it. Thanks.

> Unless @timwoj has a different opinion on this, I personally wouldn't squash since it messes with the discussions we had here.

Works for me if you want to keep those follow-up commits.

Would you be up for merging this then?

PR #261, opened 2022-08-04
Hex escape broker::to_string(data<string>) containing non-ascii chars

awelzel:
...trying something as a fix for zeek/zeek#2272 .

@Neverlord - not sure this is quite the right place and approach and you may outright reject - no hard feelings. This is similar to what @AmazingPP had done for RenderMessage(), but ensures that other `broker::to_string().c_str()` users wouldn't have `NUL` chars terminate output.

Not sure it's okay to just change the semantics here and if it's actually limited to broker::to_string. If broker::to_string() is mostly for debugging / informative messages only, it might be okay?

### DIFF
--- a/bindings/python/data.cpp
+++ b/bindings/python/data.cpp
@@ -236,7 +236,14 @@ void init_data(py::module& m) {
 	})
     .def("as_vector", [](const broker::data& d) { return broker::get<broker::vector>(d); })
     .def("get_type", &broker::data::get_type)
-    .def("__str__", [](const broker::data& d) { return broker::to_string(d); })
+    .def("__str__",
+         [](const broker::data& d) {
+           // Don't to_string a string to avoid hex escaping it.
+           if (broker::holds_alternative<std::string>(d))
+             return broker::get<std::string>(d);
+
+           return broker::to_string(d);
+         })
     .def(hash(py::self))
     .def(py::self < py::self)
     .def(py::self <= py::self)
--- a/include/broker/convert.hh
+++ b/include/broker/convert.hh
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <algorithm>
 #include <chrono>
 #include <optional>
 #include <ostream>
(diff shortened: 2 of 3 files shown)


awelzel:
Hmm, the Python binding tests failed, as the `str()` method uses `to_string()`. I ad-hoc patched this out to continue to work, but feels a bit off and maybe `__str__()` for Type.String could yield an escaped string instead.

The other users on the Zeek side appear to use `broker::to_string()` for debug messages mostly and maybe the binding should've used something (or the `str()` value should't be relied on in the tests: `broker.Data.as_string()` will return a bytes object instead.

Neverlord:
If we check the format with tests, people may already depend on Broker not escaping characters like `\t`.

> The other users on the Zeek side appear to use `broker::to_string()` for debug messages mostly

"Mostly" as in, we might still break something down the line? :)

What if we don't escape and instead print the content of the string as hex (or base64 or something like that) if it contains unprintable characters (in particular `\0`)?

awelzel:
> "Mostly" as in, we might still break something down the line? :)

This was the one that seemed non-debug, but might be fine to change as well.

https://github.com/zeek/zeek/blob/cb3abccfb19aee5ea54cbd5228d23c71202fe22e/src/broker/Data.cc#L1187

> What if we don't escape and instead print the content of the string as hex (or base64 or something like that) if it contains unprintable characters (in particular \0)?

Yep, but that would imply changing all callers of `broker::to_print(...).c_str()|.data()` to account for that?  It's what @AmazingPP had started in zeek/zeek#2272. There aren't that many callers, so maybe that's the better approach. Just seemed by those above `c_str()` examples that expectations might have been different and that's where the implied hex encoding came from.

awelzel:
>> What if we don't escape and instead print the content of the string as hex (or base64 or something like that) if it contains unprintable characters (in particular \0)?

> Why would that require updating callers? The API remains the same. Only change would be that if you have received an ill-formed string before (that contained non-printable characters) there wasn't much a caller could safely do with it whereas after the change you get at least something that's always safe to print.

Sorry - thought the implementation of "print content of the string as hex" would naively be something like `broker::maybe_hexlify(broker::to_string(data))` for the Zeek side users, but realize you meant something else.

I start to think that auditing/changing the `broker::to_string().c_str()|.data()` users on the Zeek side may be most effective after all and may just close this here again :-/

awelzel:
Closing for now.

PR #263, opened 2022-08-05
Add getters for bucket settings on family handles

### LINKED ISSUE #262: Support querying bucket count and upper bounds through metric family handles
It is currently not possible (or I couldn't figure it out) to introspect the number of buckets and their individual upper bounds using an `dbl_histogram_family_hdl` or `int_histogram_family_hdl`, but one needs to have a concrete `dbl_histogram_hdl` or `int_histogram_hdl` to gather that info.

> Seems worth creating a ticket for on the Broker-side to provide this, and then clean this up here. Shouldn't be too difficult I assume. (But ok to leave as is for now).

_Originally posted by @rsmmr in https://github.com/zeek/zeek/pull/2273#discussion_r938530690_

Neverlord:
Closes #262.

### DIFF
--- a/include/broker/telemetry/fwd.hh
+++ b/include/broker/telemetry/fwd.hh
@@ -134,6 +134,16 @@ int64_t value(const int_gauge_hdl* hdl) noexcept;
 int_gauge_hdl* int_gauge_get_or_add(metric_family_hdl* hdl,
                                     span<const label_view> labels);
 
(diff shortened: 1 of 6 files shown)


rsmmr (approved):

awelzel (approved):
Thanks! As you saw, the Zeek side changes are over here: https://github.com/zeek/zeek/pull/2325

PR #265, opened 2022-08-15
Set SO_REUSEADDR for the Prometheus HTTP server

Neverlord:
We already set SO_REUSEADDR by default for incoming Broker peerings, so omitting it for the Prometheus server is inconsistent and can be problematic when restarting an application.

### DIFF
--- a/src/endpoint.cc
+++ b/src/endpoint.cc
@@ -269,7 +269,7 @@ public:
   }
 
   expected<uint16_t> start(uint16_t port, caf::actor core,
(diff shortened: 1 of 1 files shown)


Neverlord:
@rsmmr, @timwoj: since no one gave green light yet, any objections against the change? 🙂

awelzel (approved):
FWIW, I ran into this while using the builtin Prometheus server.

rsmmr (approved):

Issue #254, opened 2022-07-09
Add additional metrics


Neverlord:
To allow users to better understand what Broker is doing and to understand the health of the system, we need additional metrics. The list we've came up with so far:

- Unsent messages, i.e., number of buffered messages at the core actor
- Unsent bytes, i.e., accumulated sizes of buffered messages
- Input and output channels of stores (a master will have one channel per clone)
- Number of entries in the data store
- Number of buffered commands, i.e., command messages that wait for one or more ACKs

Ideas for metrics that would need some extra clarification or are not straightforward to implement:

- High-water metrics, e.g., largest message yet / highest amount of buffered messages?
- Latencies: this one is tricky, because ideally this time would be measured from creating an event (in Zeek) to writing the message to a socket; however, messages currently have no timestamps

Issue #250, opened 2022-07-05
promtool check metrics inconsistencies in "internal" metrics


awelzel:
Running `promtool check metrics` for the broker / caf generated Prometheus metrics indicates some conventions are not adhered to.
```
$ curl -sf http://localhost:4242/metrics | promtool check metrics
broker_processed_elements counter metrics should have "_total" suffix
caf_system_queued_messages_total non-counter metrics should not have "_total" suffix
process_cpu_seconds_total non-counter metrics should not have "_total" suffix
```

`process_cpu_seconds_total`  could just be switched to a counter. Not sure about the semantics of `caf_system_queued_messages_total` (gauge vs counter). Might also be just a type difference.

Neverlord:
The metrics that come from CAF code (the process metrics also come from CAF) probably need to wait for CAF 0.19, the Broker metric I can update right away.

PR #266, opened 2022-08-15
Add additional health and performance metrics

Neverlord:
This PR adds a couple of new metrics to Broker that keep track of message and store command throughput as well as buffer slots. It also introduces a new metric factory to allow us to have all the strings (names and help texts) in a single file.

We've also talked about having metrics that estimate how many bytes are bound up in Broker messages. I've tried that and I couldn't find a good approach. We consider a message as buffered (received but not processed) when it is in any of the flow stages in the core. There are three stages we care about here: data inputs, command inputs and the central merge point. While messages travel through the pipeline, we convert them from data and command messages to node messages.

For data and command messages, there's no easy way of estimating the size of the payload other than applying a serializer to the messages. Even if that serializer doesn't actually write any data but only keeps track of how many bytes *would* be written: this would potentially add quite a bit of runtime overhead. Ideally, the instrumentation shouldn't add any measurable overhead to the system. If we still want that buffer size estimation in bytes, maybe we can benchmark that overhead in a real system (Zeek instance) to see whether the extra CPU load is negligible or not.

Relates #254.

### DIFF
--- a/src/internal/metric_factory.cc
+++ b/src/internal/metric_factory.cc
@@ -1,0 +1,149 @@
+#include "broker/internal/metric_factory.hh"
+
+#include <caf/actor_system.hpp>
+
+namespace broker::internal {
+
+// -- 'imports' to safe ourselves some typing ----------------------------------
+
+using dbl_gauge = metric_factory::dbl_gauge;
+
+using dbl_gauge_family = metric_factory::dbl_gauge_family;
+
+using int_gauge = metric_factory::int_gauge;
+
+using int_gauge_family = metric_factory::int_gauge_family;
+
+using dbl_counter = metric_factory::dbl_counter;
+
+using dbl_counter_family = metric_factory::dbl_counter_family;
+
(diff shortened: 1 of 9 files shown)


src/internal/metric_factory.cc
```diff
@@ -0,0 +1,149 @@
...
+  auto fm = connections_family();
+  return {
+    fm->get_or_add({{"type", "native"}}),
+    fm->get_or_add({{"type", "web-socket"}}),
+  };
+}
+
+int_counter_family* core_t::processed_messages_family() {
+  return reg_->counter_family("broker", "processed-messages", {"type"},
+                              "Total number of processed messages.", "1", true);
```
Neverlord, line 52: Note: this marks the `counter` as "total value" (passing `true` at the end) to have the exporter add a `_total` suffix for Prometheus as requested in https://github.com/zeek/broker/issues/250 (`promtool` findings).

awelzel: Thanks.

PR #274, opened 2022-08-28
Add clang-tidy to CI with an initial set of checks

Neverlord:
This PR adds `clang-tidy` to our CI pipeline with `bugprone-*` and `clang-analyzer-*` checks (plus fixes for the findings) as initial set of checks. To make the PRs easier to review, I leave it at those first two categories and enable more checks in future PRs (already working on enabling `performance-*` checks).

I've also added a CMake option (with `configure` switch) to have `clang-tidy` run during development (slows down the build quite a bit).

### DIFF
--- a/src/endpoint.cc
+++ b/src/endpoint.cc
@@ -585,7 +585,7 @@ void endpoint::shutdown() {
   BROKER_INFO("shutting down endpoint");
   if (!await_stores_on_shutdown_) {
     BROKER_DEBUG("tell core actor to terminate stores");
-    caf::anon_send(native(core_), atom::shutdown_v, atom::store_v);
+    caf::anon_send(native(core_), atom::shutdown_v, atom::data_store_v);
   }
   // Lifetime scope of the scoped actor: must go out of scope before destroying
   // the actor system.
@@ -641,7 +641,6 @@ uint16_t endpoint::listen(const std::string& address, uint16_t port,
   BROKER_INFO("try listening on"
               << (address + ":" + std::to_string(port))
               << (ctx_->cfg.options().disable_ssl ? "(no SSL)" : "(SSL)"));
-  char const* addr = address.empty() ? nullptr : address.c_str();
   uint16_t result = 0;
   caf::scoped_actor self{ctx_->sys};
   self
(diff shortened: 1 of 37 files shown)


Neverlord:
Relates #249. 🙂

src/internal/core_actor.cc
```diff
@@ -860,8 +861,11 @@ caf::error core_actor_state::init_new_peer(endpoint_id peer,
   // message size. So on the wire, every byte block that our trait produces gets
   // prefixed with 32 bit with the size.
   namespace cn = caf::net;
-  auto [rd_1, wr_1] = caf::async::make_spsc_buffer_resource<node_message>();
-  auto [rd_2, wr_2] = caf::async::make_spsc_buffer_resource<node_message>();
+  // Note: structured bindings with values confuses clang-tidy's leak checker.
+  auto resources1 = caf::async::make_spsc_buffer_resource<node_message>();
+  auto& [rd_1, wr_1] = resources1;
```
timwoj, line 866: Same question re: references. The first two changes here make the variables references, but the third one doesn't.

src/endpoint.cc
```diff
@@ -844,7 +843,9 @@ worker endpoint::do_subscribe(filter_type&& filter,
   BROKER_ASSERT(sink != nullptr);
   using caf::async::make_spsc_buffer_resource;
   // Get a pair of connected resources.
-  auto [con_res, prod_res] = make_spsc_buffer_resource<data_message>();
+  // Note: structured bindings with values confuses clang-tidy's leak checker.
+  auto resources = make_spsc_buffer_resource<data_message>();
+  auto& [con_res, prod_res] = resources;
```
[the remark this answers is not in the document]
Neverlord: Concerning the old code:

```c++
auto [con_res, prod_res] = make_spsc_buffer_resource<data_message>();
```

This deconstructs the temporary object into two values. The compiler should break this down into something like this:

```c++
auto tmp = make_spsc_buffer_resource<data_message>();
auto con_res = std::move(tmp.first);
auto prod_res = std::move(tmp.second);
```

For some reason, `clang-format` gets confused with object lifetimes for the code expanded from the structured binding and complains about a leak that doesn't happen. The new code simply makes the temporary variable visible and skips the moves by simply pointing back into the (now named) temporary object. The new version should expand to something like this:

```c++
auto resources = make_spsc_buffer_resource<data_message>();
auto& con_res = resources.first;
auto& prod_res = resources.second;
```

From a coding perspective, there is no functional change here - other than the temporary object got put on the stack and has a name now. I'd prefer the old code (simply to have a one line less of code) if it wouldn't confuse `clang-format`.

So yes, having references now is intentional to avoid extra copies. We simply need to place the result of `make_spsc_buffer_resource` on the stack for "un-confusing" `clang-format` regarding the object lifetimes.

src/internal/core_actor.cc
Neverlord: Good catch, the third one is an oversight. I'll fix it.

PR #278, opened 2022-09-11
Fix error handling with OpenSSL calls

Neverlord:
While testing Broker locally, I would sometimes run into very weird reconnect-loops in a cluster setup when restarting one of the nodes.

After much trial and error, I figured out that it only happened when using OpenSSL transport and that there actually isn't any error causing the disconnect. Just OpenSSL not clearing error states and thus Broker believing that an error happened when in reality it's just reading an error that happened at some point in the past... Solution is to call `ERR_clear_error` before pretty much any OpenSSL function, because the error queue is (for whatever reason) managed separately and reading an error from OpenSSL (what we do if something goes wrong) does not clear the error (what we expected).

### DIFF
--- a/src/internal/connector.cc
+++ b/src/internal/connector.cc
@@ -207,6 +207,7 @@ ssl_context_from_cfg(const openssl_options_ptr& cfg) {
   BROKER_DEBUG(BROKER_ARG2("authentication", cfg->authentication_enabled()));
   if (cfg->authentication_enabled()) {
     // Require valid certificates on both sides.
+    ERR_clear_error();
     if (!cfg->certificate.empty()
         && SSL_CTX_use_certificate_chain_file(ctx.get(),
                                               cfg->certificate.c_str())
@@ -237,6 +238,7 @@ ssl_context_from_cfg(const openssl_options_ptr& cfg) {
     if (SSL_CTX_set_cipher_list(ctx.get(), "HIGH:!aNULL:!MD5") != 1)
       throw ssl_error("failed to set cipher list");
   } else { // No authentication.
+    ERR_clear_error();
     SSL_CTX_set_verify(ctx.get(), SSL_VERIFY_NONE, nullptr);
 #if defined(SSL_CTX_set_ecdh_auto) && (OPENSSL_VERSION_NUMBER < 0x10100000L)
     SSL_CTX_set_ecdh_auto(ctx.get(), 1);


awelzel:
I've tested this branch with current master and it fixes the observations made in zeek/zeek#2389

0xxon:
Hi,

could you expand on the reasoning a bit more? I don't entirely get the change - even after reading the related https://www.arangodb.com/2014/07/started-hate-openssl/.

From what I understand, the problem seems to be that an error is raised - potentially in a different thread - and that that error will be returned if checking for errors, even in the current thread. I don't actually see any `openssl_get_error` happening in the code-block that is affected here. Are those happening inside the OpenSSL functions?

Also - I am concerned about the impact of threading here. Is there any chance that with this change the following will happen:

* main thread: function errors
* broker thread: error is cleared
* broker thread: openssl function call
* main thread: check if error occurred - no error returned incorrectly

Note that we potentially call a lot of OpenSSL functions in the main thread (for certificate validation in observed traffic). If this could cause wrong results there would be really bad, because it could flag invalid certificates as being valid.

0xxon:
Ok, looking at the OpenSSL source, the error state seems to be thread local - this alleviates my fears a lot (see https://github.com/openssl/openssl/blob/5639ee79bdc905c1b7ed308fcd29e515e4f01a50/crypto/err/err.c#L684). It still leaces me a bit confused, because the ArangoDB writeup made it seem like threading has something to do with this.

I am still curious where this aborts - is this inside the OpenSSL functions?

Assuming my understanding about the errors being local to threads is correct this should be ok and not impact Zeek functionality.

Neverlord:
> Ok, looking at the OpenSSL source, the error state seems to be thread local - this alleviates my fears a lot

Yeah, a single error queue for the entire process would make OpenSSL practically unusable.

In Broker, the issue came up when killing one of the peers. In that case, OpenSSL would run into an error and we'd read that error and close the socket accordingly. However, the error wouldn't go away. So at some future point in time, Broker would eventually run into EAGAIN. Then, Broker would see that the read has failed and check whether it can retry or not. Since the error is still lingering around, OpenSSL would report an error that happened way back instead of reporting EAGAIN correctly (i.e., empty SSL error queue plus `errno` set to EAGAIN).

timwoj:
In that case, would it be better to clear out the error where it occurred (after the disconnection) instead?

Neverlord:
> In that case, would it be better to clear out the error where it occurred (after the disconnection) instead?

I think clearing the error implicitly when reading/writing is much more robust and fixes this for good. Clearing an error manually is very easy to forget and we'd probably re-introduce the same bug at some point in the future again. Doing it this way is also much closer to the regular `recv` and `send` functions. You don't need to manually reset `errno` before calling either one.

timwoj (approved):

PR #283, opened 2022-09-24
Fix messaging between clone and proxy

Neverlord:
Fixes `proxy ... received an unexpected message: message(stale_data(1))` errors that started to appear with Zeek 5.0.


timwoj (approved):

timwoj:
Is this a candidate for a 5.0 patch release and should it go into 5.1?

Neverlord:
> Is this a candidate for a 5.0 patch release and should it go into 5.1?

It's a pure bug fix and appears to do what it's supposed to (see the internal ticket), so I'd say 'yes' to both. But maybe @rsmmr and @ckreibich have different opinions?

PR #285, opened 2022-09-25
Fix CMake package file when bundling CAF

Neverlord:
I noticed the other day that the CMake packaging for `libbroker` is broken ever since we switched a setup that makes CAF opaque. It's still unconditionally asking to find a CAF package, even when using the bundled version (default). This means users have to install a matching CAF release even though nothing actually requires it anymore. This matters for Zeek only when trying to use an external Broker library (which isn't officially supported, but this is how I found out that Broker ships broken CMake packages during development).

The new version turns the `find_dependecy` line off for the (default) bundled build.

A default build, after installing to a `local` prefix:

```
$ fgrep -R CAF local/lib/cmake .
local/lib/cmake/Broker/BrokerConfig.cmake:set(BROKER_USE_EXTERNAL_CAF OFF)
local/lib/cmake/Broker/BrokerConfig.cmake:if(BROKER_USE_EXTERNAL_CAF)
local/lib/cmake/Broker/BrokerConfig.cmake:  find_dependency(CAF 0.18.5 REQUIRED COMPONENTS core io net)
```

A build configured to use an external CAF, after installing to a `local` prefix:

```
$ fgrep -R CAF local/lib/cmake .
local/lib/cmake//Broker/BrokerTargets.cmake:  INTERFACE_LINK_LIBRARIES "Threads::Threads;OpenSSL::SSL;OpenSSL::Crypto;CAF::core;CAF::io;CAF::net"
local/lib/cmake//Broker/BrokerConfig.cmake:set(BROKER_USE_EXTERNAL_CAF ON)
local/lib/cmake//Broker/BrokerConfig.cmake:if(BROKER_USE_EXTERNAL_CAF)
local/lib/cmake//Broker/BrokerConfig.cmake:  find_dependency(CAF 0.18.5 REQUIRED COMPONENTS core io net)
```

That first hit is:

```
set_target_properties(broker PROPERTIES
  INTERFACE_LINK_LIBRARIES "Threads::Threads;OpenSSL::SSL;OpenSSL::Crypto;CAF::core;CAF::io;CAF::net"
)
```

Note that there's no hit in the targets file in the default build (as there shouldn't be).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -210,6 +210,7 @@ if (CAF_ROOT)
     find_package(CAF REQUIRED COMPONENTS openssl test io core)
   endif()
   list(APPEND LINK_LIBS CAF::core CAF::io)
+  set(BROKER_USE_EXTERNAL_CAF ON)
 else ()
   message(STATUS "Using bundled CAF")
   set(CAF_ENABLE_EXAMPLES OFF CACHE INTERNAL "")
@@ -219,11 +220,15 @@ else ()
   add_subdirectory(caf EXCLUDE_FROM_ALL)
   list(APPEND OPTIONAL_SRC $<TARGET_OBJECTS:libcaf_core_obj>)
   list(APPEND OPTIONAL_SRC $<TARGET_OBJECTS:libcaf_io_obj>)
+  set(BROKER_USE_EXTERNAL_CAF OFF)
 endif ()
 
 # NOTE: building and linking against a different version of the CAF incubator is
 #       also FOR DEVELOPMENT ONLY.
 if (CAFIncubator_ROOT)
+  if(NOT BROKER_USE_EXTERNAL_CAF)
+    message(FATAL_ERROR "Cannot use bundled CAF with an external incubator")
+  endif()
   if (CMAKE_VERSION VERSION_LESS 3.12)
     find_package(CAFIncubator REQUIRED COMPONENTS net PATHS "${CAF_ROOT}")
   else()
@@ -231,6 +236,9 @@ if (CAFIncubator_ROOT)
   endif()
   list(APPEND LINK_LIBS CAF::net)
 else()
+  if(BROKER_USE_EXTERNAL_CAF)
+    message(FATAL_ERROR "Cannot use an external CAF with the bundled incubator")
+  endif()
   message(STATUS "Using bundled CAF incubator")
   set(CAF_INC_ENABLE_TESTING OFF CACHE INTERNAL "")
   set(CAF_INC_ENABLE_BB_MODULE OFF CACHE INTERNAL "")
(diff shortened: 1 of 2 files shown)


timwoj (approved):

ckreibich:
Dominik, this change implies that whenever one configures an external CAF one also needs an external incubator, right? If so, I think that poses a configuration problem: when using `configure` we can currently only specify an external CAF, not an external incubator. This tripped up the Coverity action (details [here](https://github.com/zeek/zeek/actions/runs/3293864275/jobs/5442294447)):

```
CMake Error at auxil/broker/CMakeLists.txt:240 (message):
  Cannot use an external CAF with the bundled incubator
```
The fix for the Coverity build is to simply switch it back to bundled CAF, but the above still looks like a problem if we want to support external CAFs.

Neverlord:
> (...) if we want to support external CAFs.

In my mind, the purpose of shifting CAF-based Broker components to the new `internal` namespace, tightly controlling the CAF version and making it opaque at the binary level was pretty much to prevent users from plugging in a potentially unsupported CAF version.

During development, it makes things easier (at least for me) if I can checkout CAF, Broker and Zeek side-by-side (i.e., not loading the respective submodules at all) and point Broker to a local CAF version when working on a CAF bug that affects Broker and/or when working on a CAF API addition for Broker. I wouldn't deploy such a version, though. Also, I actually don't need the `configure` script option at all since I'm using CMake directly. And it's just a matter of setting `CAF_ROOT` and `CAFIncubator_ROOT`.

To cut a long story short: I think the safer option is actually stripping the `--with-caf=` option altogether. I think just by having the option, we'll signal to users / package maintainers that this is something "reasonable" to do while in reality you shouldn't use an external CAF unless you are a Broker/Zeek developer and know what you're getting yourself into.

For developers that still prefer the `configure` script over using `CMake` directly: why don't we add a way to add CMake variables directly? That could just be parameters that the `configure` script passes through without interpreting them. Something like: `./configure -D CAF_ROOT:PATH=<local-dir-1> -D CAFIncubator_ROOT:PATH=<local-dir-2>`. That would give developers full flexibility without exposing `configure` script options that users shouldn't touch. This pass-though option also avoids similar one-offs in the future. Thoughts? 🙂

ckreibich:
Ah, I see — you do want to keep the possibility of specifying alternate CAFs, just not via `configure`. I'm fine with that. I also like the generic cmake settings mechanism, because we've clearly struggled to keep explicit configure-level and cmake-level knobs consistent. Plus, it'd avoid having to replicate configure flags needed for embedded projects in higher-level ones.

ckreibich:
>  I think the safer option is actually stripping the `--with-caf=` option altogether. 

It's more severe, no? It doesn't work now (the Coverity build tried, but now failed), so we have to remove it.

PR #286, opened 2022-10-18
Adding Windows support for the Broker submodule

voidbar:


CMakeLists.txt
```diff
@@ -82,24 +82,24 @@ if (NOT broker_is_subproject)
...
+include(RequireCXX17)
 
-  # Append our extra flags to the existing value of CXXFLAGS.
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${EXTRA_FLAGS}")
+# Append our extra flags to the existing value of CXXFLAGS.
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${EXTRA_FLAGS}")
 
-endif()
-
-include(RequireCXX17)
```
timwoj, line 102: What's the reasoning for moving this `include` up before the `set` above it? I understand the change to move settings the flags outside of the `if` block they're currently in. But this doesn't seem necessary. Not a big deal if it is, I'm just trying to reduce the changeset to what's truly needed.

timwoj (changes requested):

timwoj (approved):

PR #290, opened 2022-11-05
Minimize messaging in sim_clock::advance_time

Neverlord:
This applies the performance optimization that we've discussed in https://github.com/zeek/zeek/issues/2454.

The gist of it is the sim-clock now sends its `sync_point` messages to all stores at once and then collects all responses in a second loop, also using a single timeout message instead of having one per store.

### DIFF
--- a/src/endpoint.cc
+++ b/src/endpoint.cc
@@ -207,22 +207,38 @@ public:
         --pending_count_;
       }
     }
-    // Wait for response messages.
+    // Send messages to all actors that we sync with.
     caf::scoped_actor self{ctx_->sys};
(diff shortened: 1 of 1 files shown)


awelzel (approved):
Approving based on testing it last week and still recognizing the original proposal ;-)

awelzel:
> The gist of it is the sim-clock now sends its sync_point messages to all stores at once and then collects all responses in a second loop, also using a single timeout message instead of having one per store.

It also cancels/disposes the timeout message which I still believe is the main culprit of #2454: Queueing up lots of timeout messages and re-sorting these over and over in caf.clock.

PR #291, opened 2022-11-06
Final LGTM fixes and removal

Neverlord:
Nothing fancy, just some `consider passing a const pointer/reference instead` findings. There were also some "hygiene" findings on the Python bindings that I didn't apply. I'd like to touch these parts as little as possible since we're already thinking about phasing out the bindings anyways.


Neverlord:
When merging this, it's probably also a good idea to disable LGTM in the GitHub project (i.e., remove from the "Installed GitHub Apps").

timwoj (approved):

PR #300, opened 2022-11-26
Implement graceful disconnect handshake

### LINKED ISSUE #293: Sporadic test failures on CI
We had a couple of CI runs lately that failed in the system test suite. So far, I've only seen Linux builds run into this and haven't seen it locally (macOS). So first step is trying to get this to reproduce with some frequency on a Linux image.

Neverlord:
This set of changes achieves a couple of things:

- Move all state related to a single peering into a new class to have single "point of concern".
- Add a new "handshake" for gracefully removing peers. We use the already existing ping/pong messages for this in order to avoid changing the network protocol. This also means that this is a backwards compatible change. With this new handshake, I couldn't reproduce #293 locally anymore. Before the changes, one run out of a couple dozens would fail. After the changes, I couldn't get this to reproduce in 10k runs.
- Create a single observable for peerings that "surrounds" the peer messages with the peering-related status messages. Before, these events would take a different path through the pipeline, which means peering-added events may be observed after the first message and peering-removed events before the last message. We've seen Zeek's `broker.disconnect` fail for this reason recently.

Closes #293.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -320,6 +320,7 @@ set(BROKER_SRC
   src/internal/metric_factory.cc
   src/internal/metric_scraper.cc
   src/internal/metric_view.cc
+  src/internal/peering.cc
   src/internal/pending_connection.cc
   src/internal/prometheus.cc
   src/internal/store_actor.cc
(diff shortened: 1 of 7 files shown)


Neverlord:
Looks like CI is happy. Both here and on the Zeek side: https://github.com/zeek/zeek/pull/2585. So this is ready for review now. 🙂

timwoj (approved):
I looked it over and don't see anything glaring. I can't say I understand all of it, of course.

PR #301, opened 2022-11-28
CI: add Fedora 37

ckreibich:


ckreibich:
Fedora 37 runs Python 3.11, which has changes in its Enum (sub)classes that seem to affect the tests here. The `.diag` discrepancies are all along the lines of 

```
-sent MessageType.HELLO message with 23 bytes
-received a MessageType.HELLO message with 22 bytes
+sent 6 message with 23 bytes
+received a 6 message with 22 bytes
```
which fits [changes](https://docs.python.org/3/library/enum.html#enum.IntEnum) to `enum.IntEnum`.

Neverlord (approved):

PR #306, opened 2023-01-23
sqlite_backend: Add a few more backend_options

awelzel:
* synchronous to run 'PRAGMA synchronous=<value>' during initialization.
* journal_mode to run 'PRAGMA journal_mode=<value>' during initialization.
* failure_mode to delete the database and reopen it when an error occurs during initialization.
* integrity_check to run PRAGMA integrity_check during initialization to reliably detect corrupted databases.

Also a bit more verbose logging when failing to create the initial tables or failing to create directories.

The PR zeek/zeek#2707 is using this and adds some btests switching on wal mode and the delete/fail functionality.


src/detail/sqlite_backend.cc
```diff
@@ -49,11 +49,68 @@ expected<data> from_blob(const void* buf, size_t size) {
...
+    return false;
+  }
+
+  if (value->name.rfind(prefix, 0) != 0) {
+    BROKER_ERROR("SQLite backend option '"
+                 << name << "' not starting with prefix " << prefix);
+    return false;
+  }
+
+  auto sstr = value->name.substr(prefix.size(), std::string::npos);
```
Neverlord, line 72: No reason to pass in the default explicitly.

```suggestion
  auto sstr = value->name.substr(prefix.size());
```

src/detail/sqlite_backend.cc
```diff
@@ -49,11 +49,68 @@ expected<data> from_blob(const void* buf, size_t size) {
...
+  if (allowed.count(sstr) == 0) {
+    BROKER_ERROR("SQLite backend option '" << name << "' has invalid value");
+    return false;
+  }
+
+  *result = sstr;
```
Neverlord, line 78:
```suggestion
  *result = std::move(sstr);
```

Neverlord (changes requested):
Just some notes on the implementation. The bigger question I have is: can we test this with some btests to avoid adding new, completely untested code?

awelzel:
> Just some notes on the implementation. The bigger question I have is: can we test this with some btests to avoid adding new, completely untested code?

I've updated the description with a link to the Zeek side of this. There are tests there.

src/detail/sqlite_backend.cc
```diff
@@ -98,13 +195,24 @@ struct sqlite_backend::impl {
       db = nullptr;
       return false;
     }
+
+    if (!pragma_synchronous.empty()) {
+      if (!exec_pragma("synchronous", pragma_synchronous))
+        return false;
+    }
```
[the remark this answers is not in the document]
awelzel: These were actually conscious, but can do the chaining of course.

src/detail/sqlite_backend.cc
```diff
@@ -49,11 +50,71 @@ expected<data> from_blob(const void* buf, size_t size) {
     return {ec::invalid_data};
 }
 
+// Find name in options and verify it starts with the given prefix,
+// if the prefix-stripped part of the value is found in allowed,
+// set result to the value.
+bool extract_optional_enum_option(
+  const broker::backend_options& options, const std::string& name,
+  const std::string_view& prefix,
```
Neverlord, line 58: A string view is a pointer and a size. Best practice is to pass this by value and give the compiler more room for optimization by not forcing pointer indirections.

```suggestion
  std::string_view prefix,
```

src/detail/sqlite_backend.cc
```diff
@@ -76,16 +137,58 @@ struct sqlite_backend::impl {
...
 
-  bool open(const std::string& path) {
-    BROKER_TRACE(BROKER_ARG(path));
+  // Run PRAGAMA command with an optional value.
+  //
+  // Assumes name and value have been verified previously and are safe
+  // to be formatted into SQL.
+  bool exec_pragma(const std::string& name,
+                   const std::string& value = std::string()) {
+    auto query = std::string("PRAGMA ");
```
Neverlord, line 146: Please follow Broker's coding style and general best practice (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es23-prefer-the--initializer-syntax).

```suggestion
    auto query = std::string{"PRAGMA "};
```

Neverlord (changes requested):
Just two more nitpicks. :)

Regarding the testing: Having tests in Zeek is of course a good thing. However, this catches bugs very late (only when trying to integrate a new Broker version into Zeek) and troubleshooting issues with Zeek's btests makes it much more difficult, because there are multiple layers of indirections between Broker and the actual test code.

It's ok if you don't have the cycles available, of course. Then I'll add tests at the Broker level later.

awelzel:
> > It's ok if you don't have the cycles available, of course. Then I'll add tests at the Broker level later.
> 
> Let me give it a whirl. I just found `cpp/backend.cc`. Likely you'll need to massage things around to fit well, but I can add some basic ones.

@Neverlord - I've added one into the btest directory. I don't think there was a lot of infrastructure in place so it's doing a few ad-hoc things. I'm sure you know how to cleanup things nicer, but this should provide some basic coverage without requiring to run Zeek's test suite.

Neverlord (approved):
Moved your `doctest` based tests to `btest`-based ones. LGTM now. 👍

PR #307, opened 2023-01-29
Add new status reporting feature

Neverlord:
This PR adds a new feature to Broker that allows users to get a JSON status report over HTTP. The status report gives a more detailed picture of the current state of Broker than the metrics do. For example, the status reports also contain what stores are attached and who's connected to them (or in case of a clone, who's the master).

Under the hood, this injects new "flow scopes" at some points into the data pipeline, so that we can tell how much data has gone through a given path. The core actor received a new handler that generates the status report on demand. Hence, no status report is generated unless a user explicitly asks for it. Users can ask Broker for a status report by accessing `/v1/status/json` instead of `/metrics` on the metrics port.

Finally, this adds two new `btest` checks that make sure that the reports are generated and contain the expected keys.

The reports are supposed to be a new tool for troubleshooting, but I think they should help us make Broker less of a black box in general. Of course we also need to add documentation that explain what the status reports actually show and how to interpret them. I plan to do a documentation update after merging the feature itself (and after using the reports a bit more myself).


Neverlord:
@timwoj, @ckreibich this hasn't a high priority, but I'd still appreciate feedback (or a merge). I could also walk you through the flow-scope thing if you have no idea what it's doing / how the underlying CAF API works. 🙂

src/internal/prometheus.cc
```diff
@@ -151,4 +151,211 @@ caf::behavior prometheus_actor::make_behavior() {
...
+        table tbl;
+        tbl.emplace(data{"error"s}, data{caf::to_string(what)});
+        on_status_request_cb(hdl, aid, tbl);
+      });
+  requests_[hdl].async_id = aid;
+}
+
+namespace {
+
+class jsonizer {
```
timwoj, line 198: We already have rapidjson in Zeek, would it be worth pulling it over here as well?

Neverlord: No, that wouldn't add any value in Broker. We already have `caf::json_writer` and `caf::json_reader` for interfacing with JSON and they nicely integrate with out `inspect` overloads to natively traverse custom types like `broker::data`. We use both in other places, e.g. in the [json_client](https://github.com/zeek/broker/blob/bc0205ce1fc06ddb91abb6744cb79c7eb846c23e/include/broker/internal/json_client.hh#L56).

The point of this `jsonizer` class is that it can exploit some knowledge we have about the `broker::data` we are dealing with. Tables will only have `string` keys, which means we can map them directly to JSON objects (instead of displaying them as [key-value lists](https://github.com/zeek/broker/blob/master/doc/web-socket.rst#table)). We could transform this object into something like a recursive map and then pass it to a `json_writer`, but then we would invoke lots of unnecessary heap allocations. Even if we make only shallow copies of the original data, it would still need to build the internal tree / hash map. The `jsonizer` does the translation while traversing the original data structure and writes directly into the output buffer without requiring an intermediary data structure.

Neverlord: @timwoj is this a class a blocker for you? If you think it's a no-go to have this, I'll re-implement it with `caf::json_writer`, but it's going to be less efficient since I'll have to emit an intermediary data structure. I think the performance does matter here, since we're running in the thread that does the I/O but if you think otherwise let me know.

I'd like to wrap up this PR.

timwoj: Looking at it again, I'm fine with it the way it is. I agree that there's nothing to gain from swapping it out with something else.

timwoj (approved):

PR #315, opened 2023-02-27
Fix regression in endpoint::peers

### LINKED ISSUE #309: endpoint::peers() always returns status "connected" (regression)
In the zeek/zeekctl project exists a tests/baseline collecting and printing peer statuses. It's using `Broker::peers()` to collect the status:

https://github.com/zeek/zeek/blob/fc0bfd21d5ff4730070dbc44742bd00481da8d38/scripts/policy/frameworks/control/controllee.zeek#L39-L57

This ends up actually being handled in internal/core_actor.cc (IIUC, which hard codes "connected" as peer status:
https://github.com/zeek/broker/blob/bc0205ce1fc06ddb91abb6744cb79c7eb846c23e/src/internal/core_actor.cc#L244-L251

The test case in zeekctl is currently expecting "PEERED" as status, so we regressed during some version bump here (Zeek 5.0 already returns CONNECTED), so might have been Zeek 4.2 times.

This is the involved baseline, last updated Nov, 2020 :man_shrugging:

https://github.com/zeek/zeekctl/blob/master/testing/Baseline/command.peerstatus/all.out

If "peered" would be the actual status of the remote endpoints, `endpoint::peers()` should reflect that.

Neverlord:
Fixes #309.


src/internal/core_actor.cc
```diff
@@ -245,8 +245,11 @@ caf::behavior core_actor_state::make_behavior() {
       std::vector<peer_info> result;
       for (const auto& [peer_id, state] : peers) {
         endpoint_info info{peer_id, state->addr()};
-        result.emplace_back(peer_info{std::move(info), peer_flags::remote,
-                                      peer_status::connected});
+        auto status = peer_statuses->get(peer_id);
+        if (status != peer_status::unknown) {
```
awelzel, line 249: You likely know better, but wondering a bit why it's okay to filter `unknown` here? There seems to be a peer in peers, but we decide to not return its information.

awelzel (approved):

src/internal/core_actor.cc
Neverlord: The `peer_statuses` is a shared map. If we get `unknown`, it basically means that the element has been removed in the meantime. I would need to look at the sources to see whether the connector (background task for making the handshaking) can remove items or if we do that only in the core actor at the moment.

But details aside, it's a thing that other threads may modify while we iterate over the list of locally known peers and `unknown` just means "no longer exists". 🙂

Neverlord: Oh, and after reading `unknown` here, we would eventually receive an event that cleans up the state in `peers` in case your wondering.

PR #325, opened 2023-03-29
Update CAF for GCC 13 build

Neverlord:
Pick up a CAF-side patch for building with GCC 13.


timwoj:
I wonder if we should update one of the CI hosts to GCC 13 as well. Otherwise this change doesn't have any actual coverage.

timwoj:
4d8b5358a28ac2475dec90f3e85ddbe0ab5c0e49 updates the opensuse-tumbleweed docker image to install gcc-13. It'd be good to pull that into this PR so that it gets some coverage via CI.

timwoj (approved):
Looks like the tumbleweed build is passing now.

PR #327, opened 2023-04-03
CMakeLists.txt: drop EXCLUDE_FROM_ALL

ffontaine:
Drop `EXCLUDE_FROM_ALL` from `add_subdirectory(caf)` or caf libraries such as `libcaf_core.so` won't be installed resulting in a runtime failure


Neverlord:
We have bundled CAF in a way that allows us to deviate from the official release versions by back-porting features or even pick features from unreleased versions. The one thing we *must not* do is to install this version.

The actual bug is that CMake created a shared library for CAF. We set `BUILD_SHARED_LIBS` to `OFF` right before adding CAF to the build. Could you please provide us with details on your setup to allow us to figure out why it's not building CAF as shared libraries in your case?

ffontaine:
This installation failure is raised on buildroot when building zeek 5.2.0 with `-DBUILD_SHARED_LIBS=ON`.
I'll retest it, but from reading the set documentation (https://cmake.org/cmake/help/latest/command/set.html), I assume that you should add `FORCE` to the set call here:
https://github.com/zeek/broker/blob/bc0205ce1fc06ddb91abb6744cb79c7eb846c23e/CMakeLists.txt#L219

Neverlord:
I think using force here would override the user setting. Maybe the best solution would be to not consume the library targets of CAF at all and instead use the OBJECT library targets of CAF. IIRC, CAF doesn't always generate the OBJECT library targets at the moment, so we probably would need to tweak the upstream CMake a bit. Let me check if our CMake setup for Broker works with the OBJECT libraries instead.

Neverlord:
> Let me check if our CMake setup for Broker works with the OBJECT libraries instead.

Turns out we've already did that, but the setup was buggy.

@ffontaine I have a patch ready in https://github.com/zeek/broker/pull/333 that should do the right thing now. It seemed to work for me locally, but could you double-check on your end as well?

ffontaine:
Your PR is fixing the issue, I'm closing this one

PR #328, opened 2023-04-05
Explicitly use std namespace when calling std::move

timwoj:
This fixes a bunch of -Wunqualified-std-cast-call warning with Clang 14+.

### DIFF
--- a/src/internal/clone_actor.cc
+++ b/src/internal/clone_actor.cc
@@ -25,8 +25,6 @@
 #include <chrono>
 #include <memory>
 
-using std::move;
-
 using namespace std::literals;
 
 namespace broker::internal {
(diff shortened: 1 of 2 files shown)


src/internal/connector_adapter.cc
```diff
@@ -148,7 +148,8 @@ void connector_adapter::async_connect(const network_info& addr,
   BROKER_TRACE(BROKER_ARG(addr));
   using caf::get;
   using std::move;
```
Neverlord, line 150: I think this should go now.

```suggestion
```

src/internal/connector_adapter.cc
```diff
@@ -172,7 +173,8 @@ void connector_adapter::async_listen(const std::string& host, uint16_t port,
   BROKER_TRACE(BROKER_ARG(host) << BROKER_ARG(port) << BROKER_ARG(reuse_addr));
   using caf::get;
   using std::move;
```
Neverlord, line 175: Same here.

```suggestion
```

src/internal/clone_actor.cc
Neverlord, line 1: This file also has `using std::move;` on line 28.

Neverlord (changes requested):

Neverlord (approved):

PR #343, opened 2023-05-02
Extensible event metadata 

awelzel:
This builds on #331 adding the more flexible `vector<vector<count, data>>` encoding to support more than just the network timestamp for events in the future.

The `ts()` accessor as well as the timestamp constructor stay around so no changes to zeek/zeek#2929 needed.

### DIFF
--- a/include/broker/zeek.hh
+++ b/include/broker/zeek.hh
@@ -105,14 +105,18 @@ protected:
 
 /// Support iteration with structured binding.
 class MetadataIterator {
+public:
   using iterator_category = std::forward_iterator_tag;
   using difference_type = std::ptrdiff_t;
   using value_type = std::pair<count, const data&>;
   using pointer = value_type*;
   using reference = value_type&;
 
-public:
   explicit MetadataIterator(const broker::data* ptr) noexcept : ptr_(ptr) {}
+
+  MetadataIterator(const MetadataIterator&) noexcept = default;
+
+  MetadataIterator& operator=(const MetadataIterator&) noexcept = default;
 
   value_type operator*() const {
     auto entry_ptr = get_if<vector>(*ptr_);
@@ -203,12 +207,11 @@ public:
   }
 
   MetadataWrapper metadata() const {
-    const vector* md_vec_ptr = nullptr;
-    auto ev_vec_ptr = get_if<vector>(as_vector()[2]);
-    if (ev_vec_ptr && ev_vec_ptr->size() >= 3)
-      md_vec_ptr = get_if<vector>((*ev_vec_ptr)[2]);
+    if (const auto* ev_vec_ptr = get_if<vector>(as_vector()[2]);
+        ev_vec_ptr && ev_vec_ptr->size() >= 3)
+      return MetadataWrapper{get_if<vector>((*ev_vec_ptr)[2])};
 
-    return MetadataWrapper{md_vec_ptr};
+    return MetadataWrapper{nullptr};
   }
 
   const std::optional<timestamp> ts() const {
--- a/include/broker/zeek.hh
+++ b/include/broker/zeek.hh
@@ -1,10 +1,24 @@
 #pragma once
 
+#include <cstddef>
+#include <iterator>
+#include <optional>
+
 #include "broker/data.hh"
+#include "broker/detail/assert.hh"
 
 namespace broker::zeek {
 
 const count ProtocolVersion = 1;
+
+/// Metadata attached to Zeek events.
+enum class Metadata : uint8_t {
+  NETWORK_TIMESTAMP = 1,
+  // 1 - 199 is reserved for Zeek. Otherwise free for external
+  // users to experiment with before potentially including
+  // in the reserved range.
+  USER_METADATA_START = 200,
+};
 
 /// Generic Zeek-level message.
 class Message {
@@ -96,7 +110,15 @@ public:
     : Message(Message::Type::Event, {std::move(name), std::move(args)}) {}
 
   Event(std::string name, vector args, timestamp ts)
-    : Message(Message::Type::Event, {std::move(name), std::move(args), ts}) {}
+    : Message(
+      Message::Type::Event,
+      {std::move(name), std::move(args),
+       vector{{vector{static_cast<count>(Metadata::NETWORK_TIMESTAMP), ts}}}}) {
+  }
+
+  Event(std::string name, vector args, vector metadata)
+    : Message(Message::Type::Event,
+              {std::move(name), std::move(args), std::move(metadata)}) {}
 
   Event(data msg) : Message(std::move(msg)) {}
 
@@ -108,18 +130,100 @@ public:
     return get<std::string>(get<vector>(as_vector()[2])[0]);
   }
 
-  const std::optional<timestamp> ts() const {
-    auto ev_vec = get<vector>(as_vector()[2]);
-    if (ev_vec.size() > 2)
-      return get<timestamp>(ev_vec[2]);
+  /// Support iteration with structured binding.
+  class metadata_iterator {
+    using iterator_category = std::forward_iterator_tag;
+    using difference_type = std::ptrdiff_t;
+    using value_type = std::pair<count, const data&>;
+    using pointer = value_type*;
+    using reference = value_type&;
 
-    return std::nullopt;
+  public:
+    metadata_iterator(const broker::data* ptr) : ptr_(ptr) {}
+
+    const std::pair<count, const data&> operator*() const {
+      auto entry_ptr = get_if<vector>(*ptr_);
+      BROKER_ASSERT(entry_ptr && entry_ptr->size() == 2);
+      const auto& entry = *entry_ptr;
+      return {get<count>(entry[0]), entry[1]};
+    }
+
+    metadata_iterator operator++() {
+      ++ptr_;
+      return *this;
+    }
+
+    metadata_iterator operator++(int) {
+      auto tmp = *this;
+      ++(*this);
+      return tmp;
+    }
+
+    bool operator!=(metadata_iterator& other) {
+      return ptr_ != other.ptr_;
+    }
+
+  private:
+    const broker::data* ptr_;
+  };
+
+  /// Supports iteration over metadata
+  class metadata_wrapper {
+  public:
+    metadata_wrapper(const vector* v) : v_(v) {}
+
+    metadata_iterator begin() const {
+      return metadata_iterator{v_->data()};
+    }
+
+    metadata_iterator end() const {
+      return metadata_iterator{v_->data() + v_->size()};
+    }
+
+    const data* value(Metadata key) const {
+      return value(static_cast<count>(key));
+    }
+
+    const data* value(count key) const {
+      for (const auto& [k, v] : *this) {
+        if (k == key)
+          return &v;
+      }
+      return nullptr;
+    }
+
+    const auto& as_vector() const {
+      return *v_;
+    }
+
+  private:
+    const vector* v_;
+  };
+
+  const std::optional<metadata_wrapper> metadata() const {
+    auto ret = std::optional<metadata_wrapper>();
+    auto ev_vec_ptr = get_if<vector>(as_vector()[2]);
+    if (ev_vec_ptr && ev_vec_ptr->size() >= 3) {
+      if (auto md_ptr = get_if<vector>((*ev_vec_ptr)[2]))
+        ret = std::optional<metadata_wrapper>(md_ptr);
+    }
+
+    return ret;
   }
 
-  std::optional<timestamp> ts() {
-    auto ev_vec = get<vector>(as_vector()[2]);
-    if (ev_vec.size() > 2)
-      return get<timestamp>(ev_vec[2]);
+  const data* metadata(count key) const {
+    if (const auto& md = metadata())
+      return md->value(key);
+    return nullptr;
+  }
+
+  const data* metadata(Metadata key) const {
+    return metadata(static_cast<count>(key));
+  }
+
+  const std::optional<timestamp> ts() const {
+    if (auto ts_ptr = metadata(Metadata::NETWORK_TIMESTAMP))
+      return get<timestamp>(*ts_ptr);
 
     return std::nullopt;
   }
@@ -156,22 +260,41 @@ public:
     if (!args_ptr)
       return false;
 
-    // Optional event timestamp
+    // Optional event metadata verification.
+    //
+    // Verify this is vector<vector<count, data>> and also
+    // check that the NETWORK_TIMESTAMP metadata has the
+    // right type because we know down here what to expect.
     if (v.size() > 2) {
-      auto ts_ptr = get_if<timestamp>(&v[2]);
-
-      if (!ts_ptr)
+      auto md_ptr = get_if<vector>(&v[2]);
+      if (!md_ptr)
         return false;
+
+      for (const auto& mde : *md_ptr) {
+        auto mdev_ptr = get_if<vector>(mde);
+        if (!mdev_ptr)
+          return false;
+
+        if (mdev_ptr->size() != 2)
+          return false;
+
+        const auto& mdev = *mdev_ptr;
+
+        auto mde_key_ptr = get_if<count>(mdev[0]);
+        if (!mde_key_ptr)
+          return false;
+
+        constexpr auto net_ts_key =
+          static_cast<count>(Metadata::NETWORK_TIMESTAMP);
+        if (*mde_key_ptr == net_ts_key) {
+          auto mde_val_ptr = get_if<timestamp>(mdev[1]);
+          if (!mde_val_ptr)
+            return false;
+        }
+      }
     }
 
     return true;
-  }
-
-  bool has_ts() const {
-    if (!valid())
-      return false;
-
-    return get<vector>(as_vector()[2]).size() > 2;
   }
 };
 
--- a/include/broker/zeek.hh
+++ b/include/broker/zeek.hh
@@ -95,6 +95,9 @@ public:
   Event(std::string name, vector args)
     : Message(Message::Type::Event, {std::move(name), std::move(args)}) {}
 
+  Event(std::string name, vector args, timestamp ts)
+    : Message(Message::Type::Event, {std::move(name), std::move(args), ts}) {}
+
   Event(data msg) : Message(std::move(msg)) {}
 
   const std::string& name() const {
@@ -103,6 +106,22 @@ public:
 
   std::string& name() {
     return get<std::string>(get<vector>(as_vector()[2])[0]);
+  }
+
+  const std::optional<timestamp> ts() const {
+    auto ev_vec = get<vector>(as_vector()[2]);
+    if (ev_vec.size() > 2)
+      return get<timestamp>(ev_vec[2]);
+
+    return std::nullopt;
+  }
+
+  std::optional<timestamp> ts() {
+    auto ev_vec = get<vector>(as_vector()[2]);
+    if (ev_vec.size() > 2)
+      return get<timestamp>(ev_vec[2]);
+
+    return std::nullopt;
   }
 
   const vector& args() const {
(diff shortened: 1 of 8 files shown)


include/broker/zeek.hh
```diff
@@ -105,6 +120,42 @@ public:
...
+
+  const std::optional<timestamp> ts() const {
+    if (auto ts_md = metadata(count(NETWORK_TIMESTAMP)))
+      return get<timestamp>(ts_md->get());
+
+    return std::nullopt;
+  }
+
+  bool has_ts() const {
+    return bool(ts());
```
Neverlord, line 156:
```suggestion
    return ts().has_value();
```

include/broker/zeek.hh
```diff
@@ -92,9 +94,22 @@ protected:
...
+    NETWORK_TIMESTAMP = 1,
+  };
+
   Event(std::string name, vector args)
     : Message(Message::Type::Event, {std::move(name), std::move(args)}) {}
 
+  Event(std::string name, vector args, timestamp ts)
+    : Message(Message::Type::Event,
+              {std::move(name), std::move(args),
+               vector{{vector{count(NETWORK_TIMESTAMP), ts}}}}) {}
```
Neverlord, line 107:
```suggestion
               vector{{vector{static_cast<coutn>(NETWORK_TIMESTAMP), ts}}}}) {}
```

awelzel, line 107: Fix C style cast.

include/broker/zeek.hh
```diff
@@ -105,6 +120,42 @@ public:
     return get<std::string>(get<vector>(as_vector()[2])[0]);
   }
 
+  /// A std::optional with a reference to generic event metadata.
+  const std::optional<std::reference_wrapper<const vector>> metadata() const {
+    const auto& ev_vec = get<vector>(as_vector()[2]);
```
awelzel, line 125: Add version that supports iteration.

awelzel (commented):
I've chatted with @Neverlord.

He likes the generic approach. Had much comments about the interface. Not keen on extending the Python bindings at all, but probably okay still.

I'm going to do another round and then :crossed_fingers: on merging.

awelzel:
@Neverlord - I've implemented the metadata_wrapper that supports iteration. Feel free to leave notes if that could be more concise or something seems unusual.

`metadata()` does return `std::optional<metadata_wrapper>` currently. It gives an indication if there was metadata at all. If you don't like the interface, happy to see a sketch of something nicer.

As chatted, it would be nice if we could keep the Python bits for testing.

include/broker/zeek.hh
```diff
@@ -1,11 +1,24 @@
...
 const count ProtocolVersion = 1;
 
+/// Metadata attached to Zeek events.
+enum class Metadata : uint8_t {
+  NETWORK_TIMESTAMP = 1,
+  // 1 - 199 is reserved for Zeek. Otherwise free for external
+  // users to experiment with before potentially including
+  // in the reserved range.
+  USER_METADATA_START = 200,
+};
```
awelzel, line 20: @ckreibich / @timwoj / @J-Gras   - for extra metadata attached to Zeek events, this is the convention @Neverlord / me came up with for the time being. There might be a handful things coming up or none at all. i don't expect hundreds.

On the wire metadata is actually identified by a `count`, so limiting at 255 is pretty arbitrary currently. There were thoughts of using strings instead of count for a bit of space saving for short identifiers (count takes 8 bytes), but also coming up with with a naming scheme within the metadata "zeek::event::network_time" etc, or "company::product::function::arg1", but I'd forgo this all and keep static and a bit magic numbers (we're coming from a proposal of placing a magic value at a positional place to a bit more extensibility).

IMO the 8bytes on the wire are a bit unfortunate, but it doesn't seem like the end of the world and maybe in the future broker could learn how to more efficiently encode small counts, too, if that's still possible.

Other thoughts around this?

J-Gras: Thank you for tackling that swiftly! I am good with everything that supports adding timestamps 😉  Out of curiosity: As I understand, one goal for the extensible event metadata is to allow adding new metadata without the need to touch broker. So I guess this implementation was done under the assumption that `zeek.hh` can and will be moved into Zeek, correct?

tests/python/zeek.py
```diff
@@ -69,14 +76,20 @@ def test_ping(self):
                 if i == 5:
                     expected_arg = expected_arg.encode('utf-8') + b'\x82'
 
+                ts_now = datetime.now().timestamp()
+
                 self.assertEqual(ev.name(), "ping")
+                self.assertLess(ts_now - ev.timestamp(), 1.0)
```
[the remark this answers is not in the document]
awelzel: Ah, thanks! I was hoping CI will cover whether I did the right thing here, but seems CI doesn't run them..

include/broker/zeek.hh
awelzel: Yeah, we need one place to define well-known metadata and zeek.hh seems fitting (Dominik has emphasized a few times it's wrongly placed in broker anyhow). The network timestamp will be a common use-case, so also seems fine to have convenience constructor/accessors on Event just for that.

With the proposal here it's possible to add additional metadata to an Event without modifying broker: Even if it's not defined in the `Metadata` enum, can write a Python client pushing `123: "my-metadata"` with events and if we had a generic Zeek side accessor bifs consume it there or consume it with another Python/Go client.

Neverlord: > wrongly placed in broker anyhow

I wouldn't say "wrongly". I'd say for "historic reasons" and because of the Python bindings. :)

If we drop the Python bindings, I think this header can move to Zeek.

tests/python/zeek.py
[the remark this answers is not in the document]
Neverlord: CI will only run the Python tests that are added to CTest: https://github.com/zeek/broker/blob/master/tests/CMakeLists.txt#L124.

include/broker/zeek.hh
```diff
@@ -1,11 +1,25 @@
...
```
Neverlord, line 21: Two remarks:

1) To me, it's confusing that `Metadata` is a class in Python but an enum in C++ world. It's also not describing the actual meta data but the type thereof. So maybe `MetaDataType` or `MetaDataField`?
2) The enum values are all-uppercase, which should be reserved for macro names. This header is using Zeek's coding style. However, even there, never code seems to use CamelCase for enum identifiers.

include/broker/zeek.hh
```diff
@@ -105,6 +130,104 @@ public:
     return get<std::string>(get<vector>(as_vector()[2])[0]);
   }
 
+  /// Support iteration with structured binding.
+  class metadata_iterator {
```
Neverlord, line 134: 1) This header is using Zeek's coding style, so: `MetadataIterator`.
2) Having this type inline introduces a tight coupling between the types that isn't really necessary. Moving this to namespace scope IMHO makes the individual parts better to read.

include/broker/zeek.hh
```diff
@@ -105,6 +130,104 @@ public:
...
+    using iterator_category = std::forward_iterator_tag;
+    using difference_type = std::ptrdiff_t;
+    using value_type = std::pair<count, const data&>;
+    using pointer = value_type*;
+    using reference = value_type&;
+
+  public:
+    metadata_iterator(const broker::data* ptr) : ptr_(ptr) {}
```
Neverlord, line 142:
```suggestion
    explicit metadata_iterator(const broker::data* ptr) noexcept : ptr_(ptr) {}
```

include/broker/zeek.hh
```diff
@@ -105,6 +130,104 @@ public:
...
+
+    metadata_iterator operator++() {
+      ++ptr_;
+      return *this;
+    }
+
+    metadata_iterator operator++(int) {
+      auto tmp = *this;
+      ++(*this);
+      return tmp;
```
Neverlord, line 159:
```suggestion
      return metadata_iterator{ptr_++};
```

include/broker/zeek.hh
```diff
@@ -105,6 +130,104 @@ public:
...
+    using iterator_category = std::forward_iterator_tag;
+    using difference_type = std::ptrdiff_t;
+    using value_type = std::pair<count, const data&>;
+    using pointer = value_type*;
+    using reference = value_type&;
+
+  public:
+    metadata_iterator(const broker::data* ptr) : ptr_(ptr) {}
+
+    const std::pair<count, const data&> operator*() const {
```
Neverlord, line 144: I don't know what you're intention was to add a `const` to the return type. If it wasn't a mistake: it's not doing what you think it does. 😉

Also, don't repeat yourself:

```suggestion
    value_type operator*() const {
```

include/broker/zeek.hh
```diff
@@ -105,6 +130,104 @@ public:
...
+      auto entry_ptr = get_if<vector>(*ptr_);
+      BROKER_ASSERT(entry_ptr && entry_ptr->size() == 2);
+      const auto& entry = *entry_ptr;
+      return {get<count>(entry[0]), entry[1]};
+    }
+
+    metadata_iterator operator++() {
```
Neverlord, line 151: This is a bug:

```suggestion
    metadata_iterator& operator++() noexcept {
```

include/broker/zeek.hh
```diff
@@ -105,6 +130,104 @@ public:
...
+      ++ptr_;
+      return *this;
+    }
+
+    metadata_iterator operator++(int) {
```
Neverlord, line 156:
```suggestion
    metadata_iterator operator++(int) noexcept {
```

include/broker/zeek.hh
```diff
@@ -105,6 +130,104 @@ public:
...
+      auto tmp = *this;
+      ++(*this);
+      return tmp;
+    }
+
+    bool operator!=(metadata_iterator& other) {
```
Neverlord, line 162:
```suggestion
    bool operator!=(metadata_iterator& other) noexcept {
```

include/broker/zeek.hh
```diff
@@ -105,6 +130,104 @@ public:
...
+      return ptr_ != other.ptr_;
+    }
+
+  private:
+    const broker::data* ptr_;
+  };
+
+  /// Supports iteration over metadata
+  class metadata_wrapper {
```
Neverlord, line 171: Same as with the iterator type: should use Zeek's naming convention.

include/broker/zeek.hh
```diff
@@ -105,6 +130,104 @@ public:
...
+  public:
+    metadata_wrapper(const vector* v) : v_(v) {}
+
+    metadata_iterator begin() const {
```
Neverlord, line 175: Calling this without doing something with the iterator is pretty much guaranteed to be a mistake. Hence:

```suggestion
    [[nodiscard]] metadata_iterator begin() const noexcept {
```

include/broker/zeek.hh
```diff
@@ -105,6 +130,104 @@ public:
...
+      return metadata_iterator{v_->data()};
+    }
+
+    metadata_iterator end() const {
```
Neverlord, line 179:
```suggestion
    [[nodiscard]] metadata_iterator end() const noexcept {
```

include/broker/zeek.hh
```diff
@@ -105,6 +130,104 @@ public:
...
+    }
+
+  private:
+    const broker::data* ptr_;
+  };
+
+  /// Supports iteration over metadata
+  class metadata_wrapper {
+  public:
+    metadata_wrapper(const vector* v) : v_(v) {}
```
Neverlord, line 173:
```suggestion
    explicit metadata_wrapper(const vector* v) : v_(v) {}
```

include/broker/zeek.hh
```diff
@@ -105,6 +130,104 @@ public:
...
+
+    const auto& as_vector() const {
+      return *v_;
+    }
+
+  private:
+    const vector* v_;
+  };
+
+  const std::optional<metadata_wrapper> metadata() const {
```
Neverlord, line 203: You make a point of returning `optional<metadata_wrapper>`, indicating the wrapper may not be null/empty. I can kinda see your reasoning for this, but from a user perspective, this wrapper just represents a sequence of elements. I don't think we'd ever have an actual need for distinguishing between `nullopt` and `empty` here (plus, there's also that weird `const` on a value return type again).

So, what I'd do instead is to allow constructing the wrapper from `nullptr`. The only member that has to go is `as_vector()`, which just looks like an unnecessary convenience to me (but it does lock the class into asserting a non-null `v_` member.

```suggestion
  metadata_wrapper metadata() const {
```

tests/cpp/zeek.cc
```diff
@@ -10,13 +10,74 @@
...
+  REQUIRE(ev2.valid());
   CHECK_EQUAL(ev2.name(), "test");
+  CHECK_EQUAL(ev2.ts(), std::nullopt);
   CHECK_EQUAL(ev2.args(), args);
 }
+
+TEST(event_ts) {
+  auto args = vector{1, "s", port(42, port::protocol::tcp)};
+  zeek::Event ev("test", vector(args), broker::timestamp(12s));
+  zeek::Event ev2(std::move(ev));
```
Neverlord, line 31: This test claims to be about the event timestamp. Why is it moving for no apparent reason? Just copy & paste from the `event` test? That test should also be split, looking at it.

tests/cpp/zeek.cc
```diff
@@ -10,13 +10,74 @@
...
+              broker::enum_value(std::string("hello")));
+}
+
+TEST(event_no_metadata) {
+  auto args = vector{1, "s", port(42, port::protocol::tcp)};
+  zeek::Event ev("test", vector(args));
+  zeek::Event ev2(std::move(ev));
+  REQUIRE(ev2.valid());
+  REQUIRE(!ev2.metadata());
+  REQUIRE(!ev2.metadata(1234));
```
Neverlord, line 74:
```suggestion
  CHECK(!ev2.metadata());
  CHECK(!ev2.metadata(1234));
```

tests/cpp/zeek.cc
```diff
@@ -10,13 +10,74 @@
...
+}
+
+TEST(event_ts_bad_type) {
+  auto args = vector{1, "s", port(42, port::protocol::tcp)};
+  auto ts_md = vector{static_cast<count>(1), broker::enum_value("invalid")};
+  auto md = vector{{ts_md}};
+  zeek::Event ev("test", vector(args), md);
+  REQUIRE(!ev.valid());
```
Neverlord, line 82:
```suggestion
  CHECK(!ev.valid());
```

Neverlord (changes requested):
I have to admit I've mostly skimmed over the Python implementation. Still not a big fan of adding something to an API that we have agreed to get rid of. ;)

include/broker/zeek.hh
awelzel: Yep, I'll remove it. The Python bindings were the only user, but I'll work-around it then.

>  (plus, there's also that weird const on a value return type again).

I understand what's wrong there now :-/

include/broker/zeek.hh
awelzel: I've renamed to MetaDataType and camel-cased.

I made the renamed the Python side but kept the class. No strong feelings and having the ready made `broker.Count` instances is convenient.

include/broker/zeek.hh
awelzel: > If it wasn't a mistake: it's not doing what you think it does. wink

Thanks for pointing this out. It's not the first time. I'll try to remember.

include/broker/zeek.hh
awelzel: > Yep, I'll remove it. The Python bindings were the only user, but I'll work-around it then.

I don't think you'll like the work-around - accessor to the raw vector - but the optional is gone now.

I've also removed the `metadata()` overloads from MetadataWrapper because now once can just call `ev.metadata().value(1234)` to do the same.

tests/cpp/zeek.cc
awelzel: Yeah, copy/paste. I've tried to reorder things and removed the unnecessary and repeated `args` as well.

awelzel:
Okay @Neverlord - thanks for going through. I hope I covered the things you pointed out. I've added the changes on top.

ckreibich:
I have a basic question re. the design. Does it make sense to hardwire metadata types this formally, and in Broker? To introduce new metadata, in Zeek we'd now need a patch to Broker to expand the `MetadataType` enum ... is that right? Whatever we do should at least work for plugins, but it would be nicer still to start from the script layer, to see how we could define as much as possible there — perhaps an enum that itemizes the possible metadata types and spells out the types for corresponding metadata? Basically, I wonder whether Broker needs to know about the contents of the metadata vector at all.

Regarding the Broker bindings, I'd be fine with simply leaving metadata support out of them, particularly if Jan's usecase doesn't actually need it, and we agree to ditch the bindings relatively soon (say within the 6.x cycle) — something we can soon hash out together. On the other hand, of course it's already basically implemented now...

We'll need to write up somewhere (like [here](https://docs.zeek.org/projects/broker/en/v2.3.0/web-socket.html#encoding-of-zeek-events)) what now needs to change for the existing WebSocket implementations. I admit that I care more about these (we have at least, uhm, three now?) than the Broker bindings. :-)

awelzel:
> To introduce new metadata, in Zeek we'd now need a patch to Broker to expand the MetadataType enum ... is that right? 

Can use a raw `count` without expanding the enum. Producer and consumer agreeing on the type and semantics is the part that actually matters and that might be more of a documentation than code issue - specifically for cross-language things. I mostly wanted to be nice with the enum, but now wish I maybe had just hard-coded the value instead :-)

The enum is in zeek.hh which has already been argued would be better placed in Zeek directly.

> Basically, I wonder whether Broker needs to know about the contents of the metadata vector at all.

It doesn't and I'm fairly neutral how/where/if we put (and/or expose) anything about the `1: NETWORK_TIMESTAMP` mapping into code. As above, I'd also okay dropping the whole enum for the time being and going the documentation route. We document a magic "1" in the websocket format as "event" message as well.

> Regarding the Broker bindings, I'd be fine with simply leaving metadata support out of them

Hmm, [2 weeks ago](https://github.com/zeek/broker/pull/331#issuecomment-1516972037) ago it still sounded we'd maintain them.

> I think it's the right call here to maintain the Broker bindings for a bit still. We love the WebSocket transport but it is still obscure 

> We'll need to write up somewhere (like [here](https://docs.zeek.org/projects/broker/en/v2.3.0/web-socket.html#encoding-of-zeek-events)) what now needs to change for the existing WebSocket implementations. I admit that I care more about these (we have at least, uhm, three now?) than the Broker bindings. :-)

Yep, let me give it a spin.

ckreibich:
> Hmm, 2 weeks ago it still sounded we'd maintain them.

Touché, yeah — sorry to contradict myself there. I've been thinking about this and in terms of our usual deprecation cycle it would align well if we deprecate the current bindings with Zeek 6, and then drop them with 6.1. I felt that this one cycle isn't really worth the effort, but I overlooked that even with this relatively speedy plan we'll have the bindings in the 6.0 LTS cycle still, so it was the right call to implement metadata support.

We should be able to come up with a WebSocket-powered Python package around the time of the Zeek 6 release, which would give people one release cycle to switch over. The data model is already there, we just need the actual I/O API.

Just some thoughts — we can talk it all over together.

include/broker/zeek.hh
```diff
@@ -105,6 +202,22 @@ public:
     return get<std::string>(get<vector>(as_vector()[2])[0]);
   }
 
+  MetadataWrapper metadata() const {
+    const vector* md_vec_ptr = nullptr;
+    auto ev_vec_ptr = get_if<vector>(as_vector()[2]);
+    if (ev_vec_ptr && ev_vec_ptr->size() >= 3)
+      md_vec_ptr = get_if<vector>((*ev_vec_ptr)[2]);
+
+    return MetadataWrapper{md_vec_ptr};
```
Neverlord, line 211: Having a variable that gets set somewhere along the way makes it harder to grasp the control flow. I think this would be more clear:

```suggestion
    if (auto ev_vec_ptr = get_if<vector>(as_vector()[2]);
        ev_vec_ptr && ev_vec_ptr->size() >= 3)
      return MetadataWrapper{get_if<vector>((*ev_vec_ptr)[2])};
    return MetadataWrapper{nullptr};
```

Neverlord (approved):
👍

PR #359, opened 2023-05-26
Pass -funsigned-char flag during compilation

timwoj:
Broker companion PR to https://github.com/zeek/zeek/pull/3087


ckreibich (approved):
This one pulls in a bit more on the cmake side than the bifcl and binpac PRs — just flagging in case there's anything concerning in there. (Nothing jumps out at me and I think having the cmake submodule at the same commit is better anyway.)

timwoj:
> This one pulls in a bit more on the cmake side than the bifcl and binpac PRs 

No, that's expected. I'm bumping the cmake module to the master's HEAD, which includes a few things we hadn't brought into the broker repo yet. I was planning to bump all of the submodules up to master once these have landed anyways, this just takes care of that.

PR #360, opened 2023-05-27
Pick up CAF patch for ISO 8601, add new btest

### LINKED ISSUE #346: docs/websocket: ISO 8601 discrepancies
The websocket docs say timestamps are ISO 8601. I got a bit bitten by the fact that the websocket format isn't quite as ISO 8601 or how other languages things it should look like. Possibly we should just remove the reference to ISO 8601 and specify what format is actually expected and that it'll be interpreted in UTC? There's also RFC3339 which seems a sane alternative, but not compatible with what we have currently.

---

* The subsecond part is optional according to ISO 8601, but our implementation requires it. That one would be nice to handle more graceful.

* The subsecond part apparently can be separated with a `,` comma, too and then this can also exist for non-second parts. 

* Looking a bit further, currently it's an error to provide a timezone offset or the `Z` suffix. So that probably needs documenting, too, but also brings up questions how timestamps are interpreted by the receiving side (without a specifier ISO 8601 suggests local interpretation instead of UTC)
  * 2023-04-18T12:13:15.000Z
  * 2023-04-18T12:13:15.000+02:00

The error looks something like this:
```
{'type': 'error', 'code': 'deserialization_failed', 'context': 'input #1 contained invalid data -> caf::pec::trailing_character("caf::pec::trailing_character at line 1, column 24 for input \\"2023-04-18T12:13:15.000Z\\"")'}
```

It's currently pretty hard to use default "ISO strings" generated by Python or JavaScript or `date` (which seems to one of the tools using `,` for separating subsecond parts):
```
In [7]: datetime.datetime.utcnow().astimezone(datetime.timezone.utc).isoformat()
Out[7]: '2023-05-08T10:29:28.354611+00:00'

# Below can happen if you call `utcnow()` just in the wrong moment without the replace. Using `isoformat('T', 'microseconds')` is a fix).
In [4]: datetime.datetime.utcnow().replace(microsecond=0).isoformat()
Out[4]: '2023-05-08T12:32:03'

# JavaScript (Z suffix)
> d = new Date()
2023-05-08T12:33:45.815Z
> 
> d.toJSON()
'2023-05-08T12:33:45.815Z'
> d.toISOString()
'2023-05-08T12:33:45.815Z'

# date, comma separator and timezone offset
$ date --utc --iso-8601=ns
2023-05-08T12:41:24,026668547+00:00
```

Neverlord:
Closes #346.

@awelzel in a nutshell, these changes are:

- make the fractional second part optional and support 1-9 digits (down to nanoseconds)
- support `Z` suffix
- support UTC offset via `[+-]MM:SS` suffix

When sending a timestamp without UTC offset (or `Z` suffix), Broker will parse the timestamp according to the local timezone.


Neverlord:
I though `fromisoformat` would be a safe thing to use in Python by now, but our CI still has several Python versions where that doesn't work as expected or even not at all. 😒

timwoj:
I'm going to hold off merging this one until after zeek 6.0 is out, just out of interest of keeping the broker release branch in the current state until then. We can backport it later if desired.

Neverlord:
Then let's revert this to draft for now. This also changes how Broker formats timestamps that it sends to clients, so we probably should at least have a discussion about how we document/communicate this change.

tests/btest/web-socket/timestamp.py
```diff
@@ -0,0 +1,286 @@
...
+]
+
+# tells btest we're done by writing a file
+def write_done_file():
+    with open('done', 'w') as f:
+        f.write('done')
+
+def parse_iso_timestamp(timestamp):
+    # replace 'Z' and adjust offset format from '[+-]MM:SS' to '[+-]MMSS'
+    # (note: Python versions >= 3.7 don't need this)
```
ckreibich, line 72: Following up on our earlier conversation Dominik: you can safely assume 3.7+. We bumped our minimum requirement a while back, see [here](https://docs.zeek.org/en/master/install.html#required-dependencies).

I'm about to submit a round of CI updates since Broker's setup has fallen behind compared to Zeek `master` and I believe at least the (EOL'd) Ubuntu 18.04 that's currently still configured ran Python 3.6.

timwoj:
6.0 is out, if you want to move this back to a regular PR.

tests/btest/web-socket/timestamp.py
timwoj: Unfortunately it's not that easy: https://github.com/zeek/zeek/issues/3239

timwoj:
@Neverlord Is this ready for review again?

ckreibich:
Sorry for the late follow-up here, Dominik. My takes:
> offer a new v2 endpoint with the new date rendering

Yes, I like that — we should start making use of versioning to handle format changes.

> We also need to discuss the precision of the timestamps.

For the Python bindings use-cases I've encountered so far, nanosecond precision seems pretty unimportant. I think 6 digits is sufficient.

Neverlord:
Thanks for the feedback! Closing this then. I can probably re-use (at least parts of) the tests later on, but I'll re-implement the rendering based on the new broker variants later on.

Issue #318, opened 2023-03-19
Broker variant: immutable "data views" with monotonic buffer ressources


Neverlord:
I've been thinking about performance recently. That's why I've been revisiting the radix-tree implementation for speeding up filter lookups.

But there's also room for improvement how we represent messages in memory. We've discussed memory-mappable layouts in the past. With a memory-mappable representation, we would basically read a message from the network and then simply create a sort of wrapper that decodes the bytes on demand. The downside is that creating a memory-mappable format is more complicated and requires dedicated builder APIs. While "deserializing" a value is very trivial, accessing fields in a memory-mapped data structure can come with some overhead since data must be decoded on the fly. We also would have to change our network format.

Instead of going down this road, I think there's also another option that doesn't require us to change the network format. With a monotonic buffer resource and a custom allocator, we can flatten nested data structures like `broker::data` in memory, reduce the number of heap allocations and skip any destructors (by "winking out" the entire data structure). This is the same technique that makes RapidJSON fast.

To quantify what kind of speedup we could get, I've implemented a small micro benchmark that uses regular `broker::data` and a new `shallow_data` implementation (not fully functional, just the types I've needed for the benchmark). I've picked `shallow_data`, because the original idea was that the data would also hold references into the bytes where we've deserialized from to avoid any unnecessary copying overhead. For the benchmark, it made little difference because we only have small strings.

I've picked something small to start with, so I've used a variable called `event_1` with this content: `(1, 1, (event_1, (42, test)))`. The benchmark currently only looks at how long it takes to deserialize the data:

```
----------------------------------------------------------------------
Benchmark                            Time             CPU   Iterations
----------------------------------------------------------------------
broker_data/event_1                556 ns          556 ns      1231354
broker_data/shallow_event_1        185 ns          185 ns      3826998
```

It's a small data structure, so the runtime is fast either way. However, even for this very small data structure, we have a 3x speedup. Real-world messages will be larger and when doing thousands of these per second, the performance gain adds up quickly.

I would leave `broker::data` untouched and use the new "flattened" representation for the message types. Of course there'll be faster ways to do things in the new API. We might leave `broker::data` in for convenience or eventually fade it out. In the transition phase, I think we can make the API either backwards compatible by converting to "regular" `broker::data` where needed and otherwise keep the migration overhead minimal. We wouldn't touch the network format nor the JSON representation. We can also make this transparent to the Python bindings, if we don't remove them before that.

Neverlord:
[...]

PR #367, opened 2023-06-17
Optimize metric_collector::insert_or_update lookup

### LINKED ISSUE #366: Metrics scalability performance issue in 6.0
This has been reported during Zeek 6.0 rc1/rc2 testing as manager memory growing steadily until crashing due to OOM.

Further, when manager's memory usage is growing, it does not serve requests to `curl http://localhost:9911/metrics` anymore. The connection is accepted but no response is produced. 

The following supervisor setup reproduces the issue here. It creates two counter families with 200 counter instances for each family. With 8 workers and 3 other processes, this results in enough metrics to cause manager overload.

Either increasing the number of workers or increasing `nworkers` or `ncounters` or `nfamilies` should reproduce the issue if you have a more powerful system.

Deployments with 64 or 128 workers may trigger this observation even if the number of instances per metrics is small (10 or 20) as it's multiplied by the number of endpoints if I read the code right. Let alone setups with 500+ workers (#352).

On the Zeek side we could consider removing the metrics for log streams/writers and/or event invocations for 6.0 or reduce the synchronization interval, but this seems mostly like a performance bug.

Further down the road, it may also be more efficient to query workers on-demand rather than having workers publish their metrics every second (which most of the time will be overwritten again) and the expense of a small delay.

```
$ cat my-supervisor.zeek
@load base/frameworks/cluster
@load base/frameworks/reporter

redef LogAscii::use_json = T;
redef Broker::disable_ssl = F;


redef Reporter::info_to_stderr = T;
redef Reporter::warnings_to_stderr = T;
redef Reporter::errors_to_stderr = T;

global nworkers = 8;

event zeek_init()
        {
        # print Cluster::local_node_type();
        if ( ! Supervisor::is_supervisor() )
                return;

        Broker::listen("127.0.0.1", 9999/tcp);

        local cluster: table[string] of Supervisor::ClusterEndpoint;
        cluster["manager"] = [$role=Supervisor::MANAGER, $host=127.0.0.1, $p=10000/tcp];
        cluster["proxy"] = [$role=Supervisor::PROXY, $host=127.0.0.1, $p=10001/tcp];
        cluster["logger"] = [$role=Supervisor::LOGGER, $host=127.0.0.1, $p=10002/tcp];

        local worker_port_offset = 10100;
        local i = 0;
        while ( i < nworkers )
                {
                ++i;
                local name = fmt("worker-%03d", i);
                cluster[name] = [$role=Supervisor::WORKER, $host=127.0.0.1, $p=0/tcp, $interface="lo"];
                }

        for ( n, ep in cluster )
                {
                local sn = Supervisor::NodeConfig($name=n);
                sn$cluster = cluster;
                sn$directory = n;
                sn$env = table(["ZEEK_DEFAULT_CONNECT_RETRY"] = "1");

                if ( ep?$interface )
                        sn$interface = ep$interface;

                print "starting",  sn$name;
                local res = Supervisor::create(sn);
                if ( res != "" )
                        print fmt("supervisor failed to create node '%s': %s", sn, res);
                }
        }

@if ( ! Supervisor::is_supervisor() )
@load ./telemetry.zeek
@endif


$ cat telemetry.zeek 
redef Broker::disable_ssl = T;
global update_interval: interval = 1sec;

global nfamilies = 2;
global ncounters_per_family = 200;

type Counters: record {
  f: Telemetry::CounterFamily;
  counters: vector of Telemetry::Counter;
};


global my_families: vector of Counters;
global counters: vector of Telemetry::Counter;

event update_telemetry() {
        schedule update_interval { update_telemetry() };

        for ( _, f in my_families ) {
                for ( _, c in f$counters ) {
                        Telemetry::counter_inc(c, rand(10));
                }
        }
}

event zeek_init() {
        local i = 0;
        while ( i < nfamilies ) {
                local f = Counters(
                        $f=Telemetry::register_counter_family([
                                $prefix="zeek",
                                $name="test",
                                $unit="stuff",
                                $help_text=fmt("stuff %d", i),
                                $labels=vector("label1", "label2"),
                        ]),
                        $counters=vector(),
                );
                my_families[i] = f;
                local j = 0;
                while ( j < ncounters_per_family ) {
                        local labels = vector(cat(i), cat(j));
                        f$counters += Telemetry::counter_with(f$f, labels);
                        ++j;
                }
                ++i;
        }

        schedule update_interval { update_telemetry() };

}
```

8-workers-supervisor

Neverlord:
Apply a few optimizations to the remote metric lookups:

- In `metric_collector::labels_for`: instead of assembling the list and then sorting it, we look for the insertion point once and then assemble the `vector` such that it will be sorted.
- Call `reserve` on the vectors whenever we know the size in advance in reserve space for up to 64 instances in the metric scope to avoid frequent re-allocations
- Avoid calling `find_if`: use `lower_bound` instead and keep the vector always sorted. This gives us the same lookup complexity that `std::map` has, i.e., `O(log n)`.

Relates #366.

@awelzel can you give this a spin in your setup to see if performance is good enough with this change?

### DIFF
--- a/include/broker/internal/metric_collector.hh
+++ b/include/broker/internal/metric_collector.hh
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <set>
 #include <string_view>
 #include <utility>
 #include <vector>
@@ -51,6 +52,80 @@ public:
     const caf::telemetry::metric_family* parent_;
   };
 
+  using family_ptr = std::unique_ptr<caf::telemetry::metric_family>;
+
+  using instance_ptr = std::unique_ptr<remote_metric>;
+
+  /// Predicate for checking whether a list of labels is less than another list
+  /// of labels. Automatically un-boxes `instance_ptr`.
+  struct labels_less {
+    using is_transparent = std::true_type;
+
+    /// Compares a two individual label elements.
+    template <class T1, class T2>
+    auto cmp_element(const T1& lhs, const T2& rhs) const noexcept {
+      auto cmp1 = lhs.name().compare(rhs.name());
+      return cmp1 != 0 ? cmp1 : lhs.value().compare(rhs.value());
+    }
+
+    template <class T1, class T2>
+    bool operator()(const T1& lhs, const T2& rhs) const noexcept {
+      using remote_metric = std::unique_ptr<metric_collector::remote_metric>;
+      if constexpr (std::is_same_v<T1, remote_metric>) {
+        return (*this)(lhs->labels(), rhs);
+      } else if constexpr (std::is_same_v<T2, remote_metric>) {
+        return (*this)(lhs, rhs->labels());
+      } else {
+        if (lhs.size() != rhs.size())
+          return lhs.size() < rhs.size();
+        // Empty lists are equal.
+        if (lhs.empty())
+          return false;
+        // Compare the first n - 1 labels.
+        size_t index = 0;
+        for (; index < lhs.size() - 1; ++index) {
+          auto res = cmp_element(lhs[index], rhs[index]);
+          // If a label was less than the corresponding label in the other list,
+          // we can return true immediately.
+          if (res < 0)
+            return true;
+          // If a label was greater than the corresponding label in the other
+          // list, we can return false immediately.
+          if (res > 0)
+            return false;
+          // Otherwise, the labels are equal and we continue with the next
+          // label.
+        }
+        // Compare the last label. Must be less.
+        return lhs[index] < rhs[index];
+      }
+    }
+  };
+
+  /// Predicate for checking whether two lists of labels are equal.
+  /// Automatically un-boxes `instance_ptr`.
+  struct labels_equal {
+    using is_transparent = std::true_type;
+
+    template <class T1, class T2>
+    bool operator()(const T1& lhs, const T2& rhs) const {
+      using remote_metric = std::unique_ptr<metric_collector::remote_metric>;
+      if constexpr (std::is_same_v<T1, remote_metric>) {
+        return (*this)(lhs->labels(), rhs);
+      } else if constexpr (std::is_same_v<T2, remote_metric>) {
+        return (*this)(lhs, rhs->labels());
+      } else {
+        if (lhs.size() != rhs.size())
+          return false;
+        for (size_t index = 0; index < lhs.size(); ++index) {
+          if (lhs[index] != rhs[index])
+            return false;
+        }
+        return true;
+      }
+    }
+  };
+
   // --- constructors and destructors ------------------------------------------
 
   metric_collector();
@@ -74,20 +149,11 @@ public:
 private:
   // -- private member types ---------------------------------------------------
 
-  using family_ptr = std::unique_ptr<caf::telemetry::metric_family>;
-
-  using instance_ptr = std::unique_ptr<remote_metric>;
-
   struct metric_scope {
     /// The metric family.
     family_ptr family;
     /// The instances of the metric family, sorted by label values.
-    std::vector<instance_ptr> instances;
-
-    metric_scope() {
-      // Allocate space for 64 instances to avoid frequent reallocations.
-      instances.reserve(64);
-    }
+    std::set<instance_ptr, labels_less> instances;
   };
 
   using name_map = std::unordered_map<std::string, metric_scope>;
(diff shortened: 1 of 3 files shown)


awelzel:
> @awelzel can you give this a spin in your setup to see if performance is good enough with this change?

Does not appear better. Actually based on the ratio in flamegraphs it got worse. Time is now spent in `std::vector<>::insert`.

gh-366-1

Neverlord:
Interesting. My mental model was that the insertions should happen only once, so the extra moving around in the vectors wouldn't matter. Let me update my mental model. 🙂

Neverlord:
@awelzel next try with `O(1)` insertion and `O(log n)` lookup: `std::set`.

awelzel:
I've tested the latest `set` version. It's definitely better. The manager is still at ~25% base CPU load even if there's no monitored traffic. That's with 8 workers. Find this a pretty large cost given the fact that most deployments are likely not actually using the Prometheus endpoint and/or consume telemetry.log.

We'll need to see what a production system with up to 100 workers and normal traffic ends-up looking like, I suspect this might still be detrimental.

I've increased the export interval to 3 and 5 seconds (IMO with 5 seconds we should consider a pull mechanism) and that improves things but ends-up with CPU spikes on the manager.

Here's a flamegraph, seems the bigger overhead is now unpacking incoming broker data (ignoring the huge chunk/artifact that I haven't understood yet on the right).

gh-366-2-set


[gh-366-2-set.svg.gz](https://github.com/zeek/broker/files/11779256/gh-366-2-set.svg.gz)

Neverlord:
From this flame graph, it seems like most overhead now comes from deserializing data. I think this is a strong indication that tackling https://github.com/zeek/broker/issues/318 might be the best bet to improve Broker performance overall.

awelzel:
> From this flame graph, it seems like most overhead now comes from deserializing data. I think this is a strong indication that tackling https://github.com/zeek/broker/issues/318 might be the best bet to improve Broker performance overall.

Okay, so I think the PR here is good for something we can try with 6.0-rc3. I'm going to approve.

---

I'm a bit concerned, however, having the metrics centralization enabled for 6.0 by default. Scaling up nworkers to 24 and nfamilies to 2 for the reproducer, pushes the manager to 55% CPU and broker communication over loopback is a steady ~1000KB/s. And that happens whether or not users are actually interested in telemetry or Prometheus at all. We'll see how this fares in Mike or @initconf's environments.

So we should also consider disabling the metrics centralization in a default setup again and have users opt-in to that. It's inconvenient for the default behavior, but the additional overhead when unused to me seems hard to justify.

As a flip experiment of a more prometheus'ish style setup: Running the 24 workers each with an individual metrics port and setting up prometheus-server to scrape the 24+3 processes individually once per second, its CPU usage is at ~6%. If scrape duration is increased to a more reasonable 5s, prometheus-server CPU usage goes down to ~2.5%. There's only CPU time consumed in Zeek processes when they are being scraped, too.

The on-demand request/proxy from manager to other Zeek nodes (which are well known on the Zeek layer) seems more reasonable in my view. I believe this can almost decently be implemented within Zeek script if we had an HTTP server.

awelzel (approved):

timwoj:
> I'm a bit concerned, however, having the metrics centralization enabled for 6.0 by default. Scaling up nworkers to 24 and nfamilies to 2 for the reproducer, pushes the manager to 55% CPU and broker communication over loopback is a steady ~1000KB/s. And that happens whether or not users are actually interested in telemetry or Prometheus at all. We'll see how this fares in Mike or @initconf's environments.

Can someone explain to me why we moved away from the decentralized version in the first place? Maybe this PR isn't the best place to discuss this, but the change (at least from my perspective working on os-perf2) kind of came out of nowhere and is now found to be objectively worse performance-wise.

awelzel:
> Can someone explain to me why we moved away from the decentralized version in the first place? 

This came with the telemetry framework. It's arguably nice nice if you only need to scrape a single endpoint and a well known port where you get all the metrics of your Zeek deployment. The port per process setup isn't as friendly. The proxy/on-demand model would mostly alleviate most of my concerns I think going forward.

timwoj:
> Agreed, and I like this too. It keeps all metrics available for those (few) who currently want to scrape them, and avoids the spikes that currently come with centralization. It also looks like an easy change to make, so still feasible for 6.0.

In the testing call, @initconf requested that we go ahead with implementing this and get it into 6.0 before we release RC3.

awelzel:
For reference: https://github.com/zeek/zeek/pull/3148

PR #369, opened 2023-07-01
Replace hand-rolled formatting utilities with fmt

Neverlord:
We currently have a custom `println` API in the `broker-node`. It's a one-off, but I'd actually like to use better-formatted output in the other tools and benchmarks as well. So we could shift the `println` API to some header for re-using it, but I think it's better to just use `libfmt` for the job. This is only for the benchmark programs and tools. We will not add `libfmt` as a dependency to Broker itself.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,14 @@ else()
 endif()
 unset(parent_dir)
 
+set(BROKER_LIBFMT_URL OFF
+    CACHE STRING "URL to libfmt source tarball or OFF for fetching from GitHub")
+
+# Use timestamps of extractions when using tarballs for libfmt.
+if (POLICY CMP0135)
+  cmake_policy(SET CMP0135 NEW)
+endif ()
+
 if ( MSVC )
   message(STATUS "Broker currently only supports static bulids on Windows")
   message(STATUS "Note: continue with ENABLE_STATIC_ONLY")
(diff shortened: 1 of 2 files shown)


timwoj:
I like the idea of bringing in libfmt, and have considered the same for Zeek a few times. I think the way you've done it here though won't work for people doing builds on systems without Internet access though, since FetchContent will fail to download the repo.

Neverlord:
@timwoj I've switched from fetching via Git to fetching via URL and added a cache variable to allow users to point FetchContent to a local tarball/zip. For example:

```
cmake -S . -B build -DBROKER_LIBFMT_URL=file://$HOME/Downloads/fmt-10.0.0.zip
```

This will work without internet connection, so users that do have that requirement can pre-download the sources. Or they can also just disable tools and tests for Broker, since Broker itself doesn't require libfmt. Hope that addresses your concerns?

timwoj:
> Hope that addresses your concerns?

I'm still not a huge fan. The tests and everything build by default, which means that a person building it without knowing they need to disable them is going to run into a build failure and wonder why. On top that, it kinda goes against how we've done things in the past for external projects (not that we always need to stick with that, of course). Is the concern that libfmt is just really large and bloats the repo?

Neverlord:
I just like the flexibility and ease of maintenance that `FetchContent` brings. It's basically just editing CMake to update dependencies vs. a more involved work flow with Git. I don't have that much of a preference, though.

When not using `FetchContent`, it's not clear to me what the obvious way would be, though. Some 3rd-party code is copied and bundled in a submodule (https://github.com/zeek/zeek-3rdparty). For pybind11, we maintain an extra submodule in a different part of the source tree (https://github.com/zeek/broker/tree/master/bindings/python/3rdparty).

So where would an fmt submodule go? How would you like it to be integrated?

Neverlord:
Also, if you think it's too much of a hassle to have the extra dependency, I could also pull the "mini-println" library to separate header and make it re-usable. In that case, just close the PR. 🙂

PR #370, opened 2023-07-27
Replace usage of deprecated CMake FindPythonInterp and FindPythonDev modules

timwoj:
This fixes a CMake warning with newer versions of CMake:

> CMake Warning (dev) at cmake/FindRequiredPackage.cmake:33 (find_package):
>   Policy CMP0148 is not set: The FindPythonInterp and FindPythonLibs modules
>   are removed.  Run "cmake --help-policy CMP0148" for policy details.  Use
>   the cmake_policy command to set the policy and suppress this warning.
> 
> Call Stack (most recent call first):
>   CMakeLists.txt:737 (FindRequiredPackage)
> This warning is for project developers.  Use -Wno-dev to suppress it.

The last commit forces the rebuild of a few of the docker images used by Cirrus since it seems that some of the cached images expired and they couldn't retrieve them anymore.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -446,33 +446,18 @@ endif ()
 
 # -- Bindings -----------------------------------------------------------------
 
+if(NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/bindings/python/3rdparty/pybind11/CMakeLists.txt")
+  message(WARNING "Skipping Python bindings: pybind11 submodule not available")
+  set(DISABLE_PYTHON_BINDINGS true)
+endif ()
+
 if (NOT DISABLE_PYTHON_BINDINGS)
   set(Python_ADDITIONAL_VERSIONS 3)
-  find_package(PythonInterp)
-  if (NOT PYTHONINTERP_FOUND)
+  find_package(Python 3.7 COMPONENTS Interpreter Development)
+
+  if (NOT Python_FOUND)
     message(STATUS "Skipping Python bindings: Python interpreter not found")
-  endif ()
-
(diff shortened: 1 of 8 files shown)


CMakeLists.txt
```diff
@@ -448,31 +448,22 @@ endif ()
...
-    set(PYTHONINTERP_FOUND false)
+    set(Python_FOUND false)
   endif ()
 
-  if (${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR} VERSION_LESS 3.5)
+  if (${Python_VERSION} VERSION_LESS 3.5)
     message(WARNING "Skipping Python bindings: Python 3.5 or greater required")
-    set(PYTHONINTERP_FOUND false)
+    set(Python_FOUND false)
   endif ()
```
Neverlord, line 464: I wonder if we can't do the following changes instead of messing with `Python_FOUND`:

- Check for `bindings/python/3rdparty/pybind11/CMakeLists.txt` before this block and force `DISABLE_PYTHON_BINDINGS` to `OFF` to skip the entire `find_package` block.
- Use `find_package(Python 3.5 COMPONENTS Interpreter Development)` instead of checking the version later by hand.

timwoj: Yeah, this is a problem in other places too. I wanted to talk to Robin about our `FindRequiredPackage` macro before I did that, but he agrees that we should just use the built-in functionality instead. It gives good error messages on failures.

timwoj: Fixed both of those recommendations in that last fixup.

Neverlord (approved):
LGTM. 👍

PR #372, opened 2023-08-15
Build broker bindings for older versions of Python

0xxon:
Recently (GH-370), the version of Python required to build the bindings was upped to 3.7. This is problematic, as we have older versions of Python in distributions that we still support; this lowers this version again.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -453,7 +453,7 @@ endif ()
 
 if (NOT DISABLE_PYTHON_BINDINGS)
   set(Python_ADDITIONAL_VERSIONS 3)
-  find_package(Python 3.7 COMPONENTS Interpreter Development)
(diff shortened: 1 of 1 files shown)


ckreibich:
> we have older versions of Python in distributions that we still support

Which are these, Johanna? 3.6 (and even 3.7, as of end of July) are EOL. Are we supporting distros that themselves are EOL?

timwoj:
This was a miss on reverting the change on my part, since I had tried to do it in other submodules before and reverted the others. Here's a failed build: https://cirrus-ci.com/build/5591051885346816

The culprits are:
- ubuntu18 with python 3.6.9
- opensuse leap with python 3.6.15
- centos7 with an unknown version of python, since their old version of cmake doesn't report it
- centosstream8 with python 3.6.8

PR #376, opened 2023-09-24
Avoid implicit conversions in Zeek utility classes

Neverlord:
This PR mainly refactors the API in `zeek.hh` as preparation for the upcoming `variant` switch. However, the changes also address some rather nasty API details like implicit conversions all over the place and very verbose code in our `valid()` implementations that make it often hard to follow what's actually being checked.

- to make transition to `variant` easier, the Zeek-compatibility classes provide an interface from `data_message` now
- the new `BatchBuilder` gives an interface for assembling batches without manually pulling out the `data` member from Zeek message types

### DIFF
--- a/include/broker/zeek.hh
+++ b/include/broker/zeek.hh
@@ -6,6 +6,7 @@
 
 #include "broker/data.hh"
 #include "broker/detail/assert.hh"
+#include "broker/message.hh"
 
 namespace broker::zeek {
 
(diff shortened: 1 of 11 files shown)


include/broker/zeek.hh
```diff
@@ -230,95 +230,43 @@ public:
...
 
   bool valid() const {
-    if (as_vector().size() < 3)
-      return false;
-
-    auto vp = get_if<vector>(&(as_vector()[2]));
-
-    if (!vp)
+    auto&& outer = data_.to_list();
+    if (outer.size() < 3)
```
timwoj, line 234: This is kind of a general comment across the code in this file, but having these magic indexes that you're supposed to know what's in each of them feels bad to me. Is there no way for this to be a struct of some sort?

Neverlord: Yeah, good call. It's a good opportunity to clean that up. I hope the last commit is more agreeable. 🙂

timwoj: That looks better to me 👍🏼

timwoj (approved):

PR #379, opened 2023-11-03
Fix Github edit links in top-right of HTML pages

awelzel:
These links 404'd because they omitted the file suffix, and they didn't really link to anything editable, just the raw file.

---

Beadcrumbs were relying on `suffix` which was available through sphinx_rtd_theme, but removed with readthedocs/sphinx_rtd_theme@c984d08dae206ae5f65176378e77efb852a278af.


timwoj (approved):

ckreibich:
This is the equivalent of zeek/zeek-docs#222.

PR #382, opened 2023-12-04
Pick up upstream CAF patch for flow::from_steps

Neverlord:
I've hit this recently with Zeek, so I've backported a CAF patch. After the `broker::variant` work, we should start following CAF releases again to receive patches like this automatically.


awelzel (approved):
It would be interesting to know if / how this would manifest in Zeek, but I guess we do not quite know?

Should this be backported into 6.0?

PR #384, opened 2023-12-14
CI updates

ckreibich:
Companion to zeek/zeek#3502.


.cirrus.yml
```diff
@@ -151,6 +151,14 @@ opensuse_tumbleweed_task:
   << : *CI_TEMPLATE
   << : *UNIX_ENV
 
+ubuntu23_task:
+  container:
+    # Ubuntu 23.04 EOL: Jan 2024
```
awelzel, line 148: Why not go with 23.10, released in October 2023, instead?

Seems reasonable to just bump to the latest interim release available given we're not specific in the task name either.

awelzel (approved):

.cirrus.yml
ckreibich: Thanks! That's just me not getting along very well with Ubuntu's release numbers. I'll switch to 23.10.

PR #385, opened 2023-12-17
Implement new format API for JSON output

Neverlord:
As discussed in https://github.com/zeek/broker/pull/360: we want to offer a `v2` endpoint for WebSocket clients. The new endpoint will render timestamps in UTC. This PR is prep work that re-implements the current encoding for WebSocket client in the namespace `broker::format::json::v1` and swaps out the rendering logic in `json_client` to use the new encoding logic. All of this only touches internals and should not affect clients.

Adding a `v2` endpoint is then going to be a followup.

Also contains a bugfix for [include/broker/format/txt.hh](https://github.com/zeek/broker/commit/fb7ac5c228d213a3e9d82f63122b2800e7758ec6#diff-3f522d77d7c4653065b3871ee97161a7bf4947c8487505dda4425f44811d38b6) where very large floating point numbers would miss the last character. We use the `txt` format only for internal logging, so I hope it's OK if I leave the commit here. I've noticed it only because I've used the `txt` implementation as a template.

@ckreibich I've split the work into multiple commits with a high-level description for each chunk of work. I hope this makes it easier to follow along when reviewing.


include/broker/format/json.hh
```diff
@@ -0,0 +1,324 @@
...
+}
+
+/// Renders `value` to `out` as `T` for `true` and `F` for `false`.
+template <class Policy = render_object, class T, class OutIter>
+std::enable_if_t<std::is_integral_v<T>, OutIter> encode(T value, OutIter out) {
+  using namespace std::literals;
+  if constexpr (std::is_same_v<T, bool>) {
+    return append_encoded<Policy>("boolean", value ? "true"sv : "false"sv, out);
+  } else if constexpr (std::is_same_v<T, count>) {
+    // An integer can at most have 20 digits (UINT64_MAX).
```
awelzel, line 124: Vanilla JavaScript cannot correctly parse numbers so large. Nothing to do, but that would probably be a topic if we had more browsers interact with Zeek.

This even made it into the RFC :-)

>  https://datatracker.ietf.org/doc/html/rfc7159#section-6
>
>   Note that when such software is used, numbers that are integers and
>   are in the range [-(2**53)+1, (2**53)-1] are interoperable in the
>   sense that implementations will agree exactly on their numeric
>   values.

Neverlord: Probably because JavaScript will store numbers as 64-bit floating point numbers? Probably not much we could do about that. Except sending integers as strings... 🤔

awelzel: Yeah, I don't think we should worry too much. If someone actually wants to work with numbers that large, they'll need to use BigInt anyhow and there's `json-bigint` that can parse this properly. 

---

Where I saw this first:

> In newer versions of the API, all large integer values are represented as strings by default.

https://developer.twitter.com/en/docs/twitter-ids

include/broker/format/txt.hh
```diff
@@ -51,7 +51,7 @@ OutIter encode(real value, OutIter out) {
   } else {
     std::vector<char> buf;
     buf.resize(size + 1); // +1 for the null terminator
-    std::snprintf(buf.data(), buf.size(), "%f", value);
+    size = std::snprintf(buf.data(), size + 1, "%f", value);
```
ckreibich, line 54: Nice catch.

ckreibich (approved):
LGTM, Dominik — my suggestions are all about comments or about adding comments to help readers understand what the formatting looks like. I'll leave that to you to handle as you see fit.

Apologies, once more, for the delay.

include/broker/format/json.hh
```diff
@@ -0,0 +1,324 @@
...
+/// Renders `value` to `out` in millisecond resolution.
+template <class Policy = render_object, class OutIter>
+OutIter encode(timestamp value, OutIter out) {
+  std::array<char, 32> buf;
+  auto size = encode_to_buf(value, buf);
+  return append_encoded<Policy>("timestamp", std::string_view{buf.data(), size},
+                                out);
+}
+
+/// Renders `value` to `out` in millisecond resolution.
```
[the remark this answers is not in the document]
Neverlord: I think that's a copy-and-paste mistake on my end, sorry. 😕

Neverlord:
@ckreibich thanks for the feedback! I've updated the comments with the missing examples / explanations. I hope I didn't miss something. 🙂

PR #388, opened 2024-02-15
Implement new variant type and builder API

timwoj:
This is just a CI sanity test for #318 before merging the branch.


Neverlord:
@timwoj as discussed, all squashed into one commit. I'll also update the Zeek-side PR to point to the new commit.

PR #390, opened 2024-02-22
Use the Canonical Project Structure for Broker

Neverlord:
For the Canonical Project Structure, see
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html.

With the new structure, we drop the `include` and `src` sub-directories and place the source files next to the header files under `libbroker`. Further, we place the unit tests next to the source file with `.test.cc` suffix.

Standalone tools have been placed alongside `libbroker` to the project root, whereas system/integration tests as well as benchmarks remain under `tests`, but are organized more consistently.


Neverlord:
@ckreibich most files have just been renamed. One notable exception is for the unit tests: I have dropped the manual `#define SUITE ...` from all of the tests and instead auto-generate the names via CMake. The `test.hh` header also has been renamed, so I had to change the `include` for this header in several files.

There are no functional changes. However, I did find obviously obsolete stuff like `tests/cpp/ssl.cc` (tests APIs that have been removed a long while ago) and `test/micro-benchmark/src/streaming.cc` (also uses long removed APIs). I've dropped these as part of the cleanup here since there's no point in trying to restore either one as the APIs under test simply aren't there anymore.

There are several tests and benchmarks that we currently don't build (commented out in CMake). I'll check each of those individually as a followup to see what can be restored and what might be obsolete (or became superfluous, e.g., due to a new btest).

Neverlord:
@ckreibich any ETA on this? I'd really like to get this in before doing tackling other issues (like #391) to avoid conflicts.

tests/benchmarks/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 find_package(benchmark QUIET)
 
 # add_subdirectory(cluster)
```
timwoj, line 3: Are the rest of these benchmarks coming back later? I know there's an open issue for the cluster one.

Neverlord: Yes, I plan to make a pass over them later. Some of them seem outdated, though. So probably not all of them will come back.

CMakeLists.txt
```diff
@@ -499,15 +336,14 @@ endif ()
...
-  enable_testing()
   add_subdirectory(tests)
 endif ()
 
 # -- Documentation ------------------------------------------------------------
 
-if (NOT WIN32 AND NOT BROKER_DISABLE_DOCS)
-  add_subdirectory(doc)
-endif ()
+# if (NOT WIN32 AND NOT BROKER_DISABLE_DOCS)
```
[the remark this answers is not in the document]
Neverlord: Good catch! Looks like an oversight. Should be back now.

timwoj (approved):

PR #395, opened 2024-03-17
Add API for deserializing data envelopes from JSON

### LINKED ISSUE #391: Make broker::data_envelope::deserialize() for JSON format accessible
I have found that I can parse Zeek events from arbitrary v1 binary format byte buffers with the following snippet:
```
    auto r = broker::data_envelope::deserialize(broker::endpoint_id::nil(), broker::endpoint_id::nil(), 0, "", payload,
                                                payload_size);
    if ( ! r )
        return std::nullopt;

    broker::zeek::Event ev(std::move(*r));
```

I have also found that I should be able to serialize Zeek events into v1 JSON byte buffers as follows:
```
    auto ev = broker::zeek::Event(event.HandlerName(), std::move(xs), broker::to_timestamp(event.timestamp));
    broker::format::json::v1::encode(ev.move_data(), std::back_inserter(buf));
```

Going from JSON to back to a broker::zeek::Event however isn't accessible unless I've missed something. I've found  the following place though:

https://github.com/zeek/broker/blob/861c20612e8b28811458ff9a96f091cc6de0c005/src/internal/json_client.cc#L126-L142
 
Could this be lifted into an API so that de-serialization can be re-used outside of json_client? Getting back the broker::data_message_ptr after the `de-re-serialization` step would be great.

Background is re-using the same serialization format over a different transport, which works for the binary format.

Neverlord:
Closes #391.

### DIFF
--- a/libbroker/broker/envelope.cc
+++ b/libbroker/broker/envelope.cc
@@ -7,6 +7,7 @@
 #include "broker/error.hh"
 #include "broker/expected.hh"
 #include "broker/format/bin.hh"
+#include "broker/internal/json.hh"
 #include "broker/internal/logger.hh"
 #include "broker/internal/type_id.hh"
 #include "broker/p2p_message_type.hh"
@@ -20,6 +21,9 @@
 #include <caf/byte_buffer.hpp>
 #include <caf/detail/ieee_754.hpp>
 #include <caf/detail/network_order.hpp>
+#include <caf/expected.hpp>
+#include <caf/json_object.hpp>
+#include <caf/json_value.hpp>
 
 namespace broker {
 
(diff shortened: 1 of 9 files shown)


Neverlord:
Reverting back to draft, it doesn't deserialize the topic from the data yet.

Neverlord:
Ready to go now.

libbroker/broker/envelope.cc
```diff
@@ -148,6 +152,39 @@ expected<envelope_ptr> envelope::deserialize(const std::byte* data,
...
 
+expected<envelope_ptr> envelope::deserialize_json(const char* data,
+                                                  size_t size) {
+  // Parse the JSON text into a JSON object.
+  auto val = caf::json_value::parse_shallow(std::string_view{data, size});
+  if (!val)
+    return error{ec::invalid_json};
+  auto obj = val->to_object();
+  // Type-checking.
+  if (obj.value("type").to_string() != "data-message")
```
awelzel, line 163: Hmm, I realize I'm looking at parsing one layer further in: The idea was, JSON contains only the serialized event vector without the `data-message` and `topic`.

In my use-case, "type" and "topic"  are in message headers, not within the payload and the payload isn't meant to be structured like the websocket protocol.

Let me check if the JSON serialization in #391 even does what I thought it did.

Neverlord: So you'd basically want to have a function that parses you just the `variant` from the JSON format? I'll add that later.

Neverlord: > Now, the main motivation is to "just" re-use existing encoding/serialization functionality from broker for events. If you think that API doesn't make much sense to have, feel free to push back.

No push back at all, I just didn't think of that use case yet. 🙂

I'll add functions that do the JSON (de)-serialization for `variant` as well.

Neverlord:
All done.

timwoj (approved):

timwoj:
I reviewed this a while ago, and just realized that I never approved it.

PR #412, opened 2024-05-30
Enforce TLS version 1.2 as minimum when using SSL

### LINKED ISSUE #389: Set TLS minimum proto version to 1.2 or greater.
One of our security team ran some tests and found that the Zeek broker ports were TLS 1.0 and 1.1.  After digging into the broker code (and falling for a red herring in the libcaf_openssl source), it was found that the minimum TLS version is not applied to the SSL context. If I am understanding the code properly, the right place to apply the limit would be in `./src/internal/connector.cc` at around line 209.  NOTE:  The patch file below is from the broker associated with Zeek 5.2.2, not the latest.

```
diff --git a/src/internal/connector.cc b/src/internal/connector.cc
index 1e9ae0d0..92f40be5 100644
--- a/src/internal/connector.cc
+++ b/src/internal/connector.cc
@@ -205,6 +205,7 @@ ssl_context_from_cfg(const openssl_options_ptr& cfg) {
 #endif
   auto ctx = caf::net::openssl::make_ctx(method);
   BROKER_DEBUG(BROKER_ARG2("authentication", cfg->authentication_enabled()));
+  SSL_CTX_set_min_proto_version(ctx.get(), TLS1_2_VERSION);
   if (cfg->authentication_enabled()) {
     // Require valid certificates on both sides.
     ERR_clear_error();
```

It would be great if this could be a configurable item (if it is not already), but the above change would satisfy our security team requirements.

Neverlord:
Closes #389.

### DIFF
--- a/libbroker/broker/internal/connector.cc
+++ b/libbroker/broker/internal/connector.cc
@@ -261,6 +261,8 @@ ssl_context_from_cfg(const openssl_options_ptr& cfg) {
     if (SSL_CTX_set_cipher_list(ctx.get(), cipher) != 1)
       throw ssl_error("failed to set anonymous cipher");
   }
+  // Prohibit outdated and insecure protocols.
+  SSL_CTX_set_min_proto_version(ctx.get(), TLS1_2_VERSION);
   return ctx;
 }
 


timwoj (approved):

Neverlord:
I'll hold on to merging this until I have tested this a bit more to see that it actually works as intended. On my development box, TLS < 1.2 seems to be disabled by default, so this patch doesn't change anything.

PR #417, opened 2024-07-07
Add more btests to replace "unit" tests

Neverlord:
Since we have access to `btest` in Broker for a while now, it's about time we replace some of our "unit" tests with `btest`-based tests instead. Historically, we have used a special fixture setup in Broker for tests that need multiple Broker endpoints to get around having to orchestrate multiple processes. This is really just working around the limitations we had at the time, i.e., only having access to a unit test framework but really needed to have system tests in place.

This isn't just about swapping tools, because the setup for these "unit" tests:

- uses some rather low-level CAF APIs
- requires in-depth knowledge about implementation details of the unit test framework
- forces us to ship code in Broker that only exists to make the unit tests work (will be removed as a followup)
- bypasses user-facing APIs such as `peer` and `listen`, which means we test against internal implementations details of Broker and some user-facing APIs remain untested


timwoj (approved):
I really like this. It adds a bunch of good, clear examples of using the broker API for messaging without any additional steps.

PR #419, opened 2024-07-12
Update cmake submodule to fix Zeek configure failures with FETCHCONTENT_FULLY_DISCONNECTED flag

timwoj:
This also updates the broker CI setup to use the same flag.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -161,18 +161,15 @@ alpine_task:
   << : *CI_TEMPLATE
   << : *UNIX_ENV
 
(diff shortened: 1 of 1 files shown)


ckreibich (approved):
LGTM. I saw caveats about the use of this flag over in [the docs](https://cmake.org/cmake/help/latest/module/FetchContent.html#variable:FETCHCONTENT_FULLY_DISCONNECTED) but since we'll have to see whether we keep any of this anyway, and it fixes things for the moment, I think this is the way to go for now.

PR #424, opened 2024-08-04
Replace CAF deserializers with Broker decoders

### LINKED ISSUE #161: Use safe data format for writing data to backends
As outlined in https://github.com/zeek/broker/pull/160, Broker currently serializes data using `caf::binary_serializer` and `caf::binary_deserializer` in its backends. This is unsafe, because the binary format may change between CAF releases. This means any CAF upgrade may break the persistent state of Broker.

We can either avoid binary formats in the first place, e.g., by using JSON, or we should implement a custom `serializer` and `deserializer` with a data format that Broker controls. That would also enable us to embed versioning information.

Neverlord:
When implementing the new networking layer and `broker::variant`, we have implemented encoders for the binary serialization format. However, the decoding has only been implemented for `broker::variant`.

This set of changes implements the decoding as well, meaning that Broker is now fully "self-hosted" and no longer relies on the binary serialization primitives in CAF which make no guarantees on the stability of the format.

There are two "flavors" for the decoding:
- A free `decode` function with a SAX-like interface. This function is tailored towards Broker's type system and can be used for decoding `broker::variant` and `broker::data`.
- A `decoder` class that is designed to work with the `inspect` API. We use this as a drop-in replacement for `caf::deserializer`, e.g., when decoding store commands.

Closes #161.

### DIFF
--- a/libbroker/broker/format/bin.hh
+++ b/libbroker/broker/format/bin.hh
@@ -70,8 +70,23 @@ bool read(const_byte_pointer& first, const_byte_pointer last, uint8_t& result);
 /// Reads a single 16-bit unsigned integer from a byte sequence.
 bool read(const_byte_pointer& first, const_byte_pointer last, uint16_t& result);
 
+/// Reads a single 16-bit unsigned integer from a byte sequence.
+bool read(const_byte_pointer& first, const_byte_pointer last, uint32_t& result);
+
 /// Reads a single 64-bit unsigned integer from a byte sequence.
 bool read(const_byte_pointer& first, const_byte_pointer last, uint64_t& result);
+
+/// Reads a single 8-bit signed integer from a byte sequence.
+bool read(const_byte_pointer& first, const_byte_pointer last, int8_t& result);
+
+/// Reads a single 16-bit signed integer from a byte sequence.
+bool read(const_byte_pointer& first, const_byte_pointer last, int16_t& result);
+
+/// Reads a single 32-bit signed integer from a byte sequence.
+bool read(const_byte_pointer& first, const_byte_pointer last, int32_t& result);
+
+/// Reads a single 64-bit signed integer from a byte sequence.
+bool read(const_byte_pointer& first, const_byte_pointer last, int64_t& result);
 
 /// Reads a single 64-bit floating point number from a byte sequence.
 bool read(const_byte_pointer& first, const_byte_pointer last, double& result);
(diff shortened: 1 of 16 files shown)


Neverlord:
@timwoj @awelzel ready to review, if one you'd like to pick it up. The CI error on alpine is unrelated (Cirrus fails to pull the image, apparently).

libbroker/broker/format/bin.hh
```diff
@@ -459,6 +512,399 @@ private:
...
+    case variant_tag::port: {
+      if (end - pos < 3)
+        return {false, end};
+      auto num = uint16_t{0};
+      if (!read(pos, end, num))
+        return {false, pos};
+      auto proto = uint8_t{0};
+      if (!read(pos, end, proto))
+        return {false, pos};
+      if (proto > 3) // 3 is the highest protocol number we support (ICMP).
```
ckreibich, line 595: Total nit, but if we know this is about ICMP we could just say `broker::port::protocol::icmp`.

ckreibich (approved):
LGTM Dominik, thanks. I've run Zeek's testsuites with it and they ran fine. I tried to eyeball performance via `broker-throughput` and it runs fine on this branch, but for master the server is currently throwing lots of `deserialization_failed("failed to parse data")` errors — I tried

```
$  ./broker-throughput --verbose --server :8888
```
and
```
$  ./broker-throughput --verbose -t 3 -r 1000 localhost:8888
```
... I can't recall whether this is known — do you know? We'll see performance numbers also via Zeek's CI setup when we bump Broker.

I can merge this too, if you're out of time.

Neverlord:
> but for master the server is currently throwing lots of `deserialization_failed("failed to parse data")` errors — I tried

Hm, that news to me. I don't think I'll be able to dig into that today. I'd like to look into that before we merge the PR, even if it means letting this sit for a while until I'm back. Would you be OK with that?

Neverlord:
Ok, so there's been two things. First, type 3 in `broker-throughput` was buggy (see #425). Second, the old code had detected this broken input data and complained, because it generated multiple table keys with the same value. I've added those checks to the new code as well.

ckreibich (approved):
Thanks Dominik — apologies as always for taking a while. Merging now.

ckreibich:
Just rebasing here since time has passed. I've tested Zeek's testsuite with this and it's all looking good.

PR #429, opened 2024-09-23
Disconnect slow peers and WS clients by default

Neverlord:
Adds a new maximum buffer size of peers and WebSocket clients. When reaching that limit, Broker will automatically sever the connection by default. Alternatively, users may select a DropOldest or DropNewest policy. Buffer sizes are configurable as well.


ckreibich (approved):
This all looks good to me Dominik. I'm sorry to have left this sitting for so long — I've been working with this branch throughout and am about to open a follow-up PR with a few tweaks, but let me merge this next. (I only rebased to bring the back change back to the top.)

PR #430, opened 2024-09-29
Remove obsolete operators.hh utility

Neverlord:
The utility classes in `operators.hh` overlap with `broker::detail::comparable`. There is no need to have both in the code base, so we remove the `operators.hh` header and instead use `comparable` consistently.


ckreibich (approved):
Nice, thank you!

PR #431, opened 2024-09-30
Fix bugprone and modernize clang-tidy findings

Neverlord:
Fix a few findings from clang-tidy from the `bugprone` and `modernize` categories.


ckreibich (approved):
LGTM. It'd be helpful in such cases to show the version of the tools and their concrete output, for future reference.

PR #436, opened 2024-12-04
Provide reason string in forced disconnects to allow Zeek to identify backpressure overflow, plus two fixes

ckreibich:
@Neverlord could you let me know if this looks okay to you? The reason-string support should hopefully be straightforward.

More interesting is ecb9aeaa, which I encountered while testing cluster recovery after forced disconnects: around half of the time the while-loop in `connector::run_impl()` ended up in a situation where it kept trying to read but failed, without realizing it had encountered an error. I'm simply adding a check of the requested-event poll mask, which seems to be resolving it.

I also noticed that Broker's master currently doesn't build with Zeek. I narrowed this down to a part of 4d9aa3bd, which I'm simply reverting here. It doesn't seem to have any impact on clang-tidy, but let me know if I'm missing anything.

Thanks!

### DIFF
--- a/libbroker/broker/internal/peering.cc
+++ b/libbroker/broker/internal/peering.cc
@@ -11,6 +11,8 @@
 
 #include <caf/scheduled_actor/flow.hpp>
 
+using namespace std::literals;
+
 namespace broker::internal {
 
 namespace {
@@ -94,9 +96,14 @@ public:
 
   node_message first() override {
     if (ptr_->removed()) {
+      auto msg = "removed connection to remote peer"s;
+      if (const auto& reason = ptr_->removed_reason(); !reason.empty()) {
+        msg += " (";
+        msg += reason;
+        msg += ')';
+      }
       return make_status_msg(endpoint_info{ptr_->peer_id(), ptr_->addr()},
-                             sc_constant<sc::peer_removed>(),
-                             "removed connection to remote peer");
+                             sc_constant<sc::peer_removed>(), msg.c_str());
     } else {
       return make_status_msg(endpoint_info{ptr_->peer_id(), ptr_->addr()},
                              sc_constant<sc::peer_lost>(),
@@ -119,9 +126,10 @@ void peering::on_bye_ack() {
   bye_timeout_.dispose();
 }
 
-void peering::force_disconnect() {
+void peering::force_disconnect(std::string reason) {
   if (!removed_) {
     removed_ = true;
+    removed_reason_ = std::move(reason);
   }
   on_bye_ack();
 }
@@ -129,8 +137,9 @@ void peering::force_disconnect() {
 void peering::schedule_bye_timeout(caf::scheduled_actor* self) {
   bye_timeout_.dispose();
   bye_timeout_ =
-    self->run_delayed(defaults::unpeer_timeout,
-                      [ptr = shared_from_this()] { ptr->force_disconnect(); });
+    self->run_delayed(defaults::unpeer_timeout, [ptr = shared_from_this()] {
+      ptr->force_disconnect("timeout during graceful disconnect");
+    });
 }
 
 void peering::assign_bye_token(std::array<std::byte, bye_token_size>& buf) {
--- a/libbroker/broker/internal/peering.hh
+++ b/libbroker/broker/internal/peering.hh
@@ -38,7 +38,7 @@ public:
 
   /// Forces the peering to shut down its connection without performing the BYE
   /// handshake.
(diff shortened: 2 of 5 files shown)


libbroker/broker/internal/peering.hh
```diff
@@ -63,6 +63,11 @@ public:
     return removed_;
   }
 
+  /// Returns the removal reason, which may be empty.
+  std::string removed_reason() const noexcept {
+    return removed_reason_;
+  }
```
Neverlord, line 69: This cannot be `noexcept` because it can thow bad_alloc. However, there's no reason to copy the string in the first place:

```suggestion
  const std::string& removed_reason() const noexcept {
    return removed_reason_;
  }
```

libbroker/broker/internal/peering.cc
```diff
@@ -119,9 +122,10 @@ void peering::on_bye_ack() {
   bye_timeout_.dispose();
 }
 
-void peering::force_disconnect() {
+void peering::force_disconnect(const std::string& reason) {
   if (!removed_) {
     removed_ = true;
+    removed_reason_ = reason;
```
Neverlord, line 128: Always forces copies, better:

```suggestion
void peering::force_disconnect(std::string reason) {
  if (!removed_) {
    removed_ = true;
    removed_reason_ = std::move(reason);
```

libbroker/broker/internal/peering.cc
```diff
@@ -94,9 +94,12 @@ class suffix_generator : public affix_generator {
 
   node_message first() override {
     if (ptr_->removed()) {
+      std::string msg{"removed connection to remote peer"};
+      std::string reason = ptr_->removed_reason();
+      if (!reason.empty())
+        msg += " (" + reason + ")";
```
Neverlord, line 100: This creates a lot of unnecessary temporaries (also should use `auto`).

```suggestion
      auto msg = "removed connection to remote peer"s;
      if (const auto& reason = ptr_->removed_reason(); !reason.empty()) {
        msg += " (";
        msg += reason;
        msg += ')';
      }
```

libbroker/broker/internal/peering.hh
```diff
@@ -38,7 +38,7 @@ public:
 
   /// Forces the peering to shut down its connection without performing the BYE
   /// handshake.
-  void force_disconnect();
+  void force_disconnect(const std::string& reason = "");
```
Neverlord, line 41:
```suggestion
  void force_disconnect(std::string reason);
```

Not a fan of the empty default. There are two more places where we call `force_disconnect` that would currently use the default:
- `core_actor_state::finalize_shutdown`: can pass something like "shutting down"
- `peering::schedule_bye_timeout`: can pass something like "timeout during graceful disconnect"

I think differentiating those with dedicated reasons would make sense as well.

Neverlord (changes requested):
Just a few nits.

libbroker/broker/internal/peering.hh
ckreibich: Yeah I was just trying to keep this minimal. I like your suggestions, will do. Thanks!

libbroker/broker/internal/peering.cc
```diff
@@ -119,18 +126,20 @@ void peering::on_bye_ack() {
   bye_timeout_.dispose();
 }
 
-void peering::force_disconnect() {
+void peering::force_disconnect(const std::string reason) {
```
Neverlord, line 129:
```suggestion
void peering::force_disconnect(std::string reason) {
```

Adding `const` here turns the `move` below to a copy.

libbroker/broker/internal/peering.hh
```diff
@@ -38,7 +38,7 @@ public:
 
   /// Forces the peering to shut down its connection without performing the BYE
   /// handshake.
-  void force_disconnect();
+  void force_disconnect(const std::string reason);
```
Neverlord, line 41: This is a sink argument.

```suggestion
  void force_disconnect(std::string reason);
```

Neverlord (changes requested):
Just one ill-placed `const`.

libbroker/broker/internal/peering.hh
ckreibich: D'oh, these were silly — thank you. (And, you're welcome to commit to my branches in such cases!)

ckreibich:
Rebased once more just to get it to run with latest CI updates.

timwoj (approved):
:shipit:

PR #439, opened 2024-12-05
Require Python 3.9

timwoj:
Ubuntu's python packages are weird. If you just install `python3.9` and `python3.9-dev` it doesn't install all of the files needed for CMake to figure out where `/usr/lib/python3/dist-packages` is. You have to at least install `python3-distutils`. This ends up installing `python3` so that's fine to. It's all really stupid.


ckreibich (approved):
LGTM. If you want to go a little deeper, the uses of
```
from __future__ import print_function
```
in some of the Python tests should be safe to drop now.

timwoj:
> LGTM. If you want to go a little deeper, the uses of
> 
> ```
> from __future__ import print_function
> ```
> 
> in some of the Python tests should be safe to drop now.

I added the same `ruff` linting that I added in other repos.

@Neverlord Could you look at the couple of TODOs I added in 9ad064297690596745a60b4c0ce386d9c64f298a? There's some questions I had about a couple of the tests.

PR #440, opened 2024-12-13
CI Updates

timwoj:
This fixes some build issues with recent updates:
- opensuse-leap VMs were still using old versions of gcc, so the recent cmake change to enforce newer compiler versions broke the CI builds. This PR aligns it with what we're installing on Zeek's images.
- ubuntu 20 was missing some necessary python3 files for `configure` to be able to find the dist-packages directory.


ckreibich (approved):
Interesting. 15.5 really does still give you Python 3.6 by default, even after running `zypper up`.

Btw 15.5 EOLs at the end of the month ... but we probably don't want to go through everything right now to ditch it. :-)

ckreibich (approved):

PR #443, opened 2025-01-23
Populate data_envelope with topic in json::decode()

awelzel:
Hey @Neverlord - I'm attempting to use `broker::format::json::v1::decode()` to decode Zeek events sent by external clients, but implemented outside of the Broker layer. The Broker WebSocket format defines the `"topic"` a client publishes to, to be part of the top-level JSON object.

This PR changes `decode()` to optionally support extracting the topic, too. Does this feel too ad-hoc? I realize `decode()` can decode non-events, too, so maybe adding a separate helper would be the better approach?

It's this part from the internal/json_client.cc, but it has a parsed JSON representation available:
https://github.com/zeek/broker/blob/5847b2a5458d03d56654e19b6b51a182476d36e5/libbroker/broker/internal/json_client.cc#L140-L142

### DIFF
--- a/libbroker/broker/format/json.cc
+++ b/libbroker/broker/format/json.cc
@@ -56,6 +56,12 @@ error decode(std::string_view str, variant& result) {
   if (!val)
     return error{ec::invalid_json};
   auto obj = val->to_object();
+
+  std::string_view topic = topic::reserved;
+  if (auto maybe_topic = obj.value("topic"); maybe_topic.is_string())
+    topic = std::string_view{maybe_topic.to_string().data(),
+                             maybe_topic.to_string().size()};
+
   // Try to convert the JSON structure into our binary serialization format.
   std::vector<std::byte> buf;
   buf.reserve(512); // Allocate some memory to avoid small allocations.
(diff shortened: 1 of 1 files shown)


Neverlord:
It definitely feels off to me to leak application logic into the decoding layer. What's wrong with asking the `variant` after decoding?

```c++
variant res;
... decode ...
if (auto topic = res.to_table()["topic"sv]; topic.is_string()) {
  ... topic.to_string() ...
}
```

Neverlord:
> It only looks at the data and @data-type, so the resulting variant won't have that.

Ah, that's the extra context that I needed. It's been a while, thanks for jogging my memory. 🙂

I do dislike the output parameter, though. What if instead of hard-coding `topic::reserved` as the topic, we take the topic from the JSON (if the field is present)? That would make it available in the envelope and you could navigate to it via `variant::shared_envelope()`.

awelzel:
> That would make it available in the envelope and you could navigate to it via variant::shared_envelope().

Yep, that works and looks nice! Re-pushed. Thanks!

Neverlord (approved):

PR #453, opened 2025-04-15
Extend event observer for peering/buffering updates and remove obsolete buffering metric

ckreibich:
(I'm just turning @Neverlord's branch into a PR here to get it merged.)

### DIFF
--- a/libbroker/broker/internal/core_actor.cc
+++ b/libbroker/broker/internal/core_actor.cc
@@ -20,12 +20,10 @@
 #include <caf/system_messages.hpp>
 #include <caf/unit.hpp>
 
-#include "broker/detail/assert.hh"
 #include "broker/detail/make_backend.hh"
 #include "broker/detail/prefix_matcher.hh"
 #include "broker/domain_options.hh"
 #include "broker/filter_type.hh"
-#include "broker/format/bin.hh"
 #include "broker/internal/checked.hh"
 #include "broker/internal/clone_actor.hh"
 #include "broker/internal/killswitch.hh"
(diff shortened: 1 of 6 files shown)


ckreibich:
I see that `init_new_client()` applies to [non-native clients](https://github.com/zeek/broker/blob/master/libbroker/broker/internal/core_actor.cc#L306-L316). Since we use the backpressure mechanism for these too, it seems we should hook their I/O into observers as well.

Neverlord:
> it looks to me like this would be best done via additional observer callbacks to distinguish clients vs peers

ACK, your additions make perfect sense. 👍

libbroker/broker/internal/core_actor.cc
```diff
@@ -1137,8 +1124,7 @@ caf::result<caf::actor> core_actor_state::attach_master(const std::string& name,
   auto in = self
               ->make_observable() //
               .from_resource(con2)
-              .map([this](const command_message& msg) {
-                metrics_for(packed_message_type::command).buffered->Increment();
+              .map([this](const command_message& msg) { //
```
timwoj, line 1169: Errant `//`, same below

timwoj (approved):

libbroker/broker/internal/core_actor.cc
ckreibich: I think these are clang-format "overrides". If you remove them, the layout works out quiet differently. I'll steer clear for now, but thank you for flagging. :+1:

Neverlord: Yep, those are to force `clang-format` to leave that line break alone. Otherwise, `clang-format` makes some questionable choices here that make the code harder to read IMHO. I'm sure there are ways to tweak the `clang-format` config to have it heave better in such cases but I didn't have the time yet to play around with it.

PR #465, opened 2025-06-06
Move to building everything with C++ 20

timwoj:
As part of https://github.com/zeek/zeek/issues/4465, we realized that building with Node v23 also requires building with C++20. This is a good opportunity to upgrade to this version of C++ across the board. This is the broker part of things. This will eventually require a PR in the zeek/cmake repo to be merged too, but that will come later.

### DIFF
--- a/.cmake-format.json
+++ b/.cmake-format.json
@@ -1,0 +1,101 @@
+{
+  "parse": {
+    "additional_commands": {
+      "CheckIPProto": {
+        "kwargs": {
+          "_proto": "*"
+        }
+      },
+      "CheckType": {
+        "kwargs": {
+          "_type": "*",
+          "_alt_type": "*",
+          "_var": "*"
+        }
+      },
+      "SetPackageVersion": {
+        "kwargs": {
+          "_version": "*"
+        }
+      },
+      "SetPackageFileName": {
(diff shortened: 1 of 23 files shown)


Neverlord:
@timwoj I have a patch ready to fix deprecation warnings from CAF headers with C++20. I can apply that to our fork, but we should pull in #461 first. That, or we need to start managing multiple branches in our CAF fork, which adds complexity.

[patch.txt](https://github.com/user-attachments/files/20729385/patch.txt)

timwoj:
@Neverlord I'm also debating splitting out the other commits in this PR except one for the CMake changes, since all of them should work without needing C++20. Do you think that makes sense? I feel like that gets us to a place where we can merge all of the various CMake-change PRs at once without merging any other functional changes. Would that work for the CAF patch too?

timwoj:
This is chopped down to just the change required to enable C++20 now. Everything else went into master early in the linked PRs.

timwoj:
I cleaned up the Windows build files a little bit, re-ran the build a few times, and now it seems happy again.

timwoj:
There's still some problems with the linker on Windows, I suppose. I'll dig into those more locally.

PR #469, opened 2025-06-24
Remove GCC 8 workaround for std::less

timwoj:
This pulls a commit out of https://github.com/zeek/broker/pull/465 that doesn't necessarily need C++ 20 to build.


Neverlord (approved):
👍

PR #474, opened 2025-07-06
Read timestamps as UTC by default

### LINKED ISSUE #434: Timestamps produced by broker::format::json::v1::encode are TZ dependent
This is related to #346 , but I wonder if for now we should enforce UTC?

The following snippet encodes the `0` timestamp, which IMO should lead to `1970-01-01 00:00.000` regardless of the environment.

```
#include <broker/format/json.hh>
#include <chrono>
#include <string>

int main() {
    std::string s;
    const auto inserter = std::back_inserter(s);
    broker::format::json::v1::encode(broker::timestamp{std::chrono::seconds::zero()}, inserter);
    std::fprintf(stderr, "%s\n", s.c_str());
}
```

Currently, the output of this program depends on the setting of `TZ`.

```
$ ./test 
{"@data-type":"timestamp","data":"1970-01-01T01:00:00.000"}
$ TZ=UTC ./test
{"@data-type":"timestamp","data":"1970-01-01T00:00:00.000"}
$ TZ=Asia/Tokyo ./test
{"@data-type":"timestamp","data":"1970-01-01T09:00:00.000"}
$ TZ=America/Los_Angeles ./test
{"@data-type":"timestamp","data":"1969-12-31T16:00:00.000"}
```

This means that two instances of broker with different TZ settings in their environment will not agree on the timestamps.

Before #346 is fixed, should we make sure to always use UTC instead of localtime?
```
--- a/libbroker/broker/format/json.cc
+++ b/libbroker/broker/format/json.cc
@@ -31,7 +31,7 @@ size_t encode_to_buf(timestamp value, std::array<char, 32>& buf) {
 #ifdef _MSC_VER
   localtime_s(&time_buf, &secs);
 #else
-  localtime_r(&secs, &time_buf);
+  gmtime_r(&secs, &time_buf);
 #endif
   buf[0] = '"';
   auto pos = strftime(buf.data() + 1, buf.size() - 1, "%FT%T", &time_buf) + 1;
```

Neverlord:
- Backport the `caf::chrono` utilities for `caf::chrono::datetime`
- Render Broker timestamps as UTC by default in the JSON encoder
- Parse timestamps in UTC by default in the JSON decoder

As a side effect of using `caf::chrono::datetime`, we also get support for `Z` suffix as well as for explicit offset syntax (e.g. `+02:00`).

Closes #434.

### DIFF
--- a/libbroker/broker/format/json.cc
+++ b/libbroker/broker/format/json.cc
@@ -29,9 +29,9 @@ size_t encode_to_buf(timestamp value, std::array<char, 32>& buf) {
   // more than enough space.
   tm time_buf;
 #ifdef _MSC_VER
-  localtime_s(&time_buf, &secs);
+  gmtime_s(&time_buf, &secs);
 #else
-  localtime_r(&secs, &time_buf);
+  gmtime_r(&secs, &time_buf);
 #endif
   buf[0] = '"';
   auto pos = strftime(buf.data() + 1, buf.size() - 1, "%FT%T", &time_buf) + 1;
--- a/libbroker/broker/format/json.test.cc
+++ b/libbroker/broker/format/json.test.cc
@@ -41,18 +41,7 @@ void strip_whitespaces(std::string& str) {
 // Returns a `time_t` value for 2014-07-09T10:16:44, i.e., the date and time of
 // the first commit to the Broker repository.
 timestamp broker_genesis() {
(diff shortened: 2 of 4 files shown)


awelzel (approved):
LGTM - I opened the submodule bump and I'm a bit overwhelmed, but treating it just as a lib/dep is ok IMO. Thanks!

PR #477, opened 2025-07-14
Get rid of `monotonic_buffer_resource`

Neverlord:
Seems like the STL version is now widely available. Hence, we no longer need our own backport version.

No functional change, just replacing hand-rolled classes with their new STL equivalent.


Neverlord:
> This looks fine to me, even though it's still in draft state.

It's basically done, but I left it in Draft state because I had a test failing on CI for the first run. Not sure if it's related or not, so I've wanted to give this more test runs.

PR #479, opened 2025-07-22
Use std::format in Broker

Neverlord:
- re-implement the logging functions to use `std::format`
- provide formatters for Broker types implementing `convert`
- explicitly generate `to_string` instead forcing them

In particular the latter change is targeting the over-reliance on "automagic" injection of functions which makes the code both brittle and hard to follow. Consequently, the `std::formatter` implementations for Broker are all opt-in via explicitly calling
`BROKER_STD_FORMATTER_IMPL`. We could try specializing based on the existence of a `convert` function, but this could pull in unwanted types and cause ambiguity. Instantiating the formatters via macro is more verbose, but also safer.

### DIFF
--- a/doc/formatting.rst
+++ b/doc/formatting.rst
@@ -1,0 +1,71 @@
+Formatting
+==========
+
+Broker uses C++20's ``std::format`` library for string formatting throughout the
+codebase and provides formatters for its own types.
+
+String Conversion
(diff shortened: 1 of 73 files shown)


ckreibich:
@timwoj this is just like what you noticed in Zeek — patchy availability of the library on the various platforms — right? We should probably just close this one for now, if so.

PR #482, opened 2025-07-24
Replace SFINAE traits with concepts

Neverlord:
I've made a pass through our `type_traits.hh` header and replaced the old, convoluted SFINAE traits with much more readable `concept` code. 🙂


ckreibich (approved):
Thumbs up from me too — sorry for the delay on my end, Dominik.

PR #485, opened 2025-07-25
Remove obsolete macro utilities

Neverlord:
BROKER_TRY is no longer used in Broker. Neither are the macro utilities that it relies on.


timwoj (approved):
I love any change that deletes a macro.

PR #489, opened 2025-11-23
Make broker-node rate calculation more accurate

Neverlord:
Decouple the message processing from the rate calculation by pushing the latter to a separate thread. By performing the rate calculation separately, we get more accurate results and also keep the code more structured.


ckreibich (approved):
LGTM Dominik, sorry to take forever here.

PR #502, opened 2026-05-17
Remove support for WebSocket clients

Neverlord:
Zeek uses its own WebSocket server (see https://github.com/zeek/zeek/pull/5464), so Broker's implementation can go now.


ckreibich (approved):
Thanks @Neverlord! Looks good to me.

PR #508, opened 2026-07-21
Add GitHub CI setup based on the Cirrus configuration

### LINKED ISSUE #507: Bring back basic CI after Cirrus shutdown
We should get basic CI back for Broker and drop the `.cirrus.yml` configuration. Github should suffice for this, and we could mirror the approach we now have [for Zeek](https://github.com/zeek/zeek/blob/master/.github/workflows/check.yml). I don't think we'd need to test as many distros, maybe just Fedora / Ubuntu / MacOS / Windows / ASan.

@awelzel, @Neverlord, fyi — thoughts?

timwoj:
This PR adds a basic GitHub CI setup. It skips using our Dockerfiles, instead adding some scripts that do the package installations and whatnot. It also removes the FreeBSD builds.

This was almost entirely generated by Claude, based on the existing Cirrus configuration. The only manual changes I made were to modify the `uses:` lines to use commit hashes instead of version tags.

Fixes #507


awelzel (approved):
LGTM.

ci/debian-12/prepare.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+set -e
+set -x
+
+export DEBIAN_FRONTEND=noninteractive
+export TZ=America/Los_Angeles
```
awelzel, line 6: I find setting TZ, specifically to America/Los_Angeles, a bit strange, but seems we're doing that over in zeek/zeek as well 🤷‍♂️

